Issue #32, opened 2020-10-09
Defer log string formatting until it's confirmed that logging is enabled


jamessimone:
Don't chew through CPU resources unnecessarily if logging at that level is not enabled.

jongpie:
Completed in #33

PR #33, opened 2020-10-11
Defer log entry formatting when logging is not enabled

jamessimone:
* Addresses #32 by short-circuiting when logging is not enabled. 
* Started adding LogEntryBuilder tests


jongpie (approved):

jongpie:
Closes #32

PR #45, opened 2020-10-14
Flow usage improvements

jamessimone:
Added quality of life improvements for Flow usage - categorization, plus ability to retrieve latest created log via Flow

### DIFF
--- a/manifest/package.xml
+++ b/manifest/package.xml
@@ -8,6 +8,7 @@
         <members>ComponentLogEntry_Tests</members>
         <members>FlowLogEntry</members>
         <members>FlowLogEntry_Tests</members>
+        <members>FlowLogRepo</members>
         <members>LogBatchPurger</members>
         <members>LogBatchPurger_Tests</members>
         <members>LogEntryBuilder</members>
--- a/manifest/package.xml
+++ b/manifest/package.xml
@@ -9,6 +9,7 @@
         <members>FlowLogEntry</members>
         <members>FlowLogEntry_Tests</members>
         <members>FlowLogRepo</members>
+        <members>FlowLogRepo_Tests</members>
         <members>LogBatchPurger</members>
         <members>LogBatchPurger_Tests</members>
         <members>LogEntryBuilder</members>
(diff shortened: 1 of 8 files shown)


nebula-logger/main/logger/classes/FlowLogRepo.cls
```diff
@@ -0,0 +1,18 @@
...
+
+    /**
+     * Returns the latest Log created to the Flow so that you can redirect a user to the Log record when required
+     *
+     * @param  `List<String>` - should be populated with the single Transaction Id for the Log you'd like to retrieve the Id for
+     * @return `List<Log__c>` - due to current Flow limitations, we return a list, but it will only ever have one Log__c entry in it
+     */
+    @InvocableMethod(category='Logging' label='Retrieve Log Id by Transaction Id' description='Returns the created log. Please only pass one Transaction Id at a time')
+    public static List<Log__c> getLogIds(List<String> transactionId) {
+        return [SELECT Id FROM Log__c WHERE TransactionId__c = :transactionId LIMIT 1];
```
jongpie, line 16: What if they call this in flow before saving any logs? In that scenario, this query will fail. Should we handle that somehow? Either return null, or save-and-then-return-log?

nebula-logger/main/logger/classes/FlowLogRepo.cls
```diff
@@ -0,0 +1,18 @@
+/*************************************************************************************************
+* This file is part of the Nebula Logger project, released under the MIT License.                *
+* See LICENSE file or go to https://github.com/jongpie/NebulaLogger for full license details.    *
+*************************************************************************************************/
+
+public without sharing class FlowLogRepo {
```
jongpie, line 6: Would you mind also adding a dedicated test class (`FlowRepo_Tests`) for this class?

nebula-logger/main/logger/classes/FlowLogRepo.cls
jamessimone: Jonathan and I spoke offline about this, and concluded: remove parameters, and just return null if the log hasn't been saved

manifest/package.xml
```diff
@@ -8,6 +8,7 @@
         <members>ComponentLogEntry_Tests</members>
         <members>FlowLogEntry</members>
         <members>FlowLogEntry_Tests</members>
+        <members>FlowLogRepo</members>
```
jongpie, line 11: Last change - can you add the new class `FlowLogRepo_Tests` to package.xml?

jongpie (approved):
Looks good - approved!

PR #51, opened 2020-10-15
Added better support for SObject record parameters in Flow

### LINKED ISSUE #46: Find a workaround for the Salesforce-required SObject type input on flows
Currently, Salesforce requires you to select an SObject type in Flows for the SObject parameter `FlowLogEntry.record`, even though it's an optional parameter. The easy workaround is to just choose a random SObject when you're not using the parameter, but need to explore if there's a better way to handle this.

jongpie:
Fixed #46 - the new Flow functionality to provide SObject record parameters (PR #43) works, but due to how Salesforce has implemented this feature, Flow always requires admins to select an SObject for the parameter (even though it's optional).

To avoid this annoyance, and because SObject parameters for Flow are still relatively new, I've removed `FlowLogEntry.record` and instead added a new class for passing SObject records, `FlowRecordLogEntry`. 
- When this action is called from Flow, the SObject parameter is required.
- `FlowLogEntry` still provides the optional parameter `FlowLogEntry.recordId` for situations where passing just an ID is sufficient (or easier)

screenshot

### DIFF
--- a/nebula-logger/main/logger/classes/FlowRecordLogEntry.cls
+++ b/nebula-logger/main/logger/classes/FlowRecordLogEntry.cls
@@ -25,7 +25,7 @@ public without sharing class FlowRecordLogEntry {
     @InvocableVariable(required=false label='(Optional) Timestamp')
     public DateTime timestamp = System.now();
 
-    @InvocableMethod(category='Logging' label='Add Log Entry for an SObject record' description='Creates a log entry a flow or process builder and stores the record as JSON')
+    @InvocableMethod(category='Logging' label='Add Log Entry for an SObject Record' description='Creates a log entry a flow or process builder and stores the record as JSON')
     public static List<String> addFlowRecordEntries(List<FlowRecordLogEntry> flowRecordLogEntries) {
         Boolean saveLog = false;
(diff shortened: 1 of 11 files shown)


nebula-logger/main/logger/classes/FlowRecordLogEntry.cls
```diff
@@ -0,0 +1,56 @@
...
+    @InvocableVariable(required=false label='(Optional) Logging Level')
+    public String loggingLevelName = 'DEBUG';
+
+    @InvocableVariable(required=false label='(Optional) Topics')
+    public List<String> topics;
+
+    @InvocableVariable(required=false label='(Optional) Timestamp')
+    public DateTime timestamp = System.now();
+
+    @InvocableMethod(category='Logging' label='Add Log Entry for an SObject Record' description='Creates a log entry a flow or process builder and stores the record as JSON')
```
jamessimone, line 28: Creates a log entry *for* a flow or process builder ...

jongpie: Fixed!

jamessimone:
@jongpie I'm just gonna test this one quick locally, but I think the changes look great :)

jamessimone:
Validated the changes locally - looks good to me!

Issue #41, opened 2020-10-14
Upgrade to dynamic forms for flexipages


jongpie:
It'd be helpful to hide/show certain fields as needes on Log and Log Entry - some fields are not always set (exception fields on log entry arent set for diagnostic entries, etc).

https://releasenotes.docs.salesforce.com/en-us/winter21/release-notes/rn_forcecom_lab_dynamic_forms_ga.htm

jongpie:
Closed by #41

PR #44, opened 2020-10-14
Add new Log__c fields for org, community and user license details

jongpie:
Closes #39 and introduces several new fields on `Log__c`

Field Label | Field Name
--- | --- 
Community Network ID | `Log__c.NetworkId__c` 
Community Name | `Log__c.NetworkName__c`
Community Login URL | `Log__c.NetworkLoginUrl__c`
Community Logout URL | `Log__c.NetworkLogoutUrl__c`
Community Self Registration URL | `Log__c.NetworkSelfRegistrationUrl__c`
Community URL Path Prefix | `Log__c.NetworkUrlPathPrefix__c`
Organization ID | `Log__c.OrganizationId__c`
Organization Name | `Log__c.OrganizationName__c`
User License Definition Key | `Log__c.UserLicenseDefinitionKey__c`
User License ID | `Log__c.UserLicenseId__c`
User License Name | `Log__c.UserLicenseName__c`

PR #55, opened 2020-10-15
Dynamic forms

jamessimone:
Addresses #41. Can't yet delete page layouts because at least one per object is required.

_It's just a biiiiiiig old blob of XML chaaaaanges_


nebula-logger/main/logger/objects/LogEntry__c/fields/HasException__c.field-meta.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jongpie, line 1: I just remembered that we might be able to do the conditional rendering in the flexipage without a new field - you could use the picklist field `LogEntry__c.Type__c == 'Exception'` to get essentially the same result. But let me know if you think it's still worth keeping the new field.

jamessimone: I think we should keep it to make things consistent in terms of how the flexipage is filtered - reduces the cognitive overhead.

jongpie: Works for me!

jongpie:
I deployed it to my sandbox to review - I think this looks good! Couple of thoughts -
- Feel free to merge (since it's being merged with your `winter21` branch instead of `master`)
- After this gets merged to `master`, I'll probably do another round of updates on the flexipages to add some of the new fields from #44 and #57 
- For now, I'm planning to maintain both the old page layout & the new dynamic flexipage layout - I'd like to keep everything working in Classic UI for a bit longer (which doesn't use flexipages at all). At some point, I might drop support for Classic UI/stop worrying about the page layout, but I'll keep it updated for the time being

jamessimone:
Sounds good, thanks man!

PR #57, opened 2020-10-16
Added 'log retention date' field for batch purge job

### LINKED ISSUE #56: Add the ability to exclude some records in the batch purge job
`LogBatchPurger` currently supports filtering records, but it's based solely on the log's created date. This works fine if you just want to delete any logs that are older than X days ago.

But there are plenty of scenarios where you may want to keep certain logs longer - due to audit requirements, ongoing troubleshooting, and so on. So there should be some way to further control which records are included/excluded in... _THE PURGE_

jongpie:
Fixed #56 by adding the field `Log__c.LogRetentionDate__c` to control when the batch class `LogBatchPurger` will delete a `Log__c` record
- The field default value is `TODAY() + 7` - so out of the box, logs will be deleted 7 days after they're generated (assuming you run the `LogBatchPurger` job)
- Admins & developers can control the deletion by simply changing the value of `Log__c.LogRetentionDate__c`
  - This approach lets admins & devs go wild with how they want to control this - the field can be set manually or via any preferred automation (Flow, Apex, API, etc)
  - To **prevent** a log record from being deleted: set `Log__c.LogRetentionDate__c` to null
  - To **postpone** a log record being deleted: set `Log__c.LogRetentionDate__c` to further in the future


jamessimone:
@jongpie I love the changes, but I do have one question - do you think it's necessary to have individual logs for each batch deleted? It might make more sense to have them as log entries within the same log that's already been created.

jongpie:
@jamessimone yeah, fair question - in an older version of this batch job, I had a project where logs with a lot of log entries would fail to delete due to a 'cascading deletion exception, cannot delete related records' (I'm paraphrasing, I don't remember the exact error now). 
- I'm not 100% sure that it truly helps to delete the log entries directly, but I don't think it hurts either - and we can always change it later, since it's just the internal logic in the batch job
- On my previous project, reducing the batch size helped with the exception, but it seeeeemed like also including the `LogEntry__c` in the delete statement (instead of letting the platform's cascade deletes handle it) helped reduce/eliminate the issue while using a larger batch size

nebula-logger/main/logger/classes/LogBatchPurger.cls
```diff
@@ -3,36 +3,53 @@
...
-        List<SObject> recordsToDelete = new List<SObject>();
-
-        for (Log__c log : scope) {
-            recordsToDelete.addAll(log.LogEntries__r);
-            recordsToDelete.add(log);
+        this.totalProcessedRecords += scope.size();
+
+        try {
+            Logger.setParentLogTransactionId(this.originalTransactionId);
+            Logger.info(new LogMessage('Starting deletion of {0} records', scope.size()));
```
jamessimone, line 31: @jongpie this was the line that I was asking about, as well as line 44 - I think there was some confusion about what I was referring to, I get why you're doing the "cascade delete" !

jongpie: Ooooooooooooooooooooooooooooooooooooooooooooh, now I see - sorry for the confusion! 

So, now that I'm looking at the right thing - it's definitely not needed, and I might make it configurable in the future (or remove them), but I have found it to be useful to be able to go back and verify that the logs are being deleted. If you're not totally against having it in there, I'll probably leave it for now

jamessimone: Haha great, got it. OK, looks good to me!

Issue #50, opened 2020-10-15
Release a managed package!


jongpie:

jongpie:
After a few months of enhancing, cleaning up and stabilizing the codebase, as well as evaluating the pros & cons of 1st gen vs 2nd gen packages, I've finally created a public beta version of a 2nd gen managed package

**Beta 2GP Managed Package**
- This is a beta version of a managed package - so it won't be upgradeable due to Salesforce limitations with beta packages 
- The package has not yet been submitted to be added to AppExchange (but I hope to do that in the near future)
- The package uses the namespace `Nebula`
- The 'View JSON' component on `Log__c` has been temporarily removed - it will be replaced with a new lwc version in issue #54 
- Package Version is 1.3.0-4 - the package version number will be updated to align with the Github relase numbers for the first non-beta release
  - Production install URL: https://login.salesforce.com/packaging/installPackage.apexp?p0=04t4x000000lnVyAAI 
  - Sandbox install URL: https://test.salesforce.com/packaging/installPackage.apexp?p0=04t4x000000lnVyAAI

PR #64, opened 2020-10-18
Updated Apex access modifiers 

jongpie:
- Converted several classes, methods & variables to `global` instead of `public` - this will be needed in order to create a managed package (#50)
- Converted Flow timestamp attributes  to be `private` so that the logging system always handles generating timestamps


.travis.yml
```diff
@@ -24,4 +24,4 @@ before_install:
 
 script:
     - sfdx force:config:set apiVersion=50.0
-    - sfdx force:source:deploy --checkonly --manifest ./manifest/package.xml --testlevel RunLocalTests --targetusername nebula_ci --verbose
+    - sfdx force:source:deploy --checkonly --sourcepath ./nebula-logger --testlevel RunLocalTests --targetusername nebula_ci --verbose
```
jamessimone, line 27: 👌👌👌

PR #68, opened 2020-10-19
Refactored how 'related record name' is set

### LINKED ISSUE #67: Convert 'related record name' formula field to text
The field `LogEntry__c.RelatedRecordName__c` is a formula fields that currently relies on lookup fields to be created, as well as the formula to be updated. This approach has 2 drawbacks
- The formula field cannot be updated when using a managed package
- The design is not scaleable because of formula limits - orgs that want to log multiple objects would need multiple lookups & a large formula

Instead, the name field of the record (typically 'Name', but there are exceptions like 'CaseNumber', 'Subject', 'Title', etc) should be stored directly on `LogEntry__c`. This will _also_ provide a snapshot of what the record's name was at the time that the log entry was created.

jongpie:
Fixes #67 - by replacing the formula field version of `LogEntry__c.RelatedRecordName__c` with a text version that's set via triggers
 - All lookup fields (RelatedAccount__c, RelatedLead__c, etc) have been deleted
    - The new approach no longer relies on lookup fields for handling the related record name
 - `LogEntryHandler` now dynamically queries for the related record names
    - This relies on querying (potentially multiple objects) to get the record names - safeguards have been added to avoid hitting the SOQL limit for the transaction


jongpie:
For future reference, the script below can be used as a demo of the functionality - it covers several standard objects with non-typical display name fields
- `Case` object uses `Case.CaseNumber`
- `CampaignMember` object does not have a display name
 - `Task` object uses `Task.Subject`
 - Included several other standard objects for good measure

```
Logger.getTransactionId();

List<SObject> records = new List<SObject>();
records.addAll([select Id, CreatedById FROM Account limit 10]);
records.addAll([select Id, CreatedById FROM Campaign limit 10]);
records.addAll([select Id, CreatedById FROM CampaignMember limit 10]);
records.addAll([select Id, CreatedById FROM Case limit 10]);
records.addAll([select Id, CreatedById FROM Contact limit 10]);
records.addAll([select Id, CreatedById FROM Lead limit 10]);
records.addAll([select Id, CreatedById FROM Opportunity limit 10]);
records.addAll([select Id, CreatedById FROM Quote limit 10]);
records.addAll([select Id, CreatedById FROM Task limit 10]);
records.addAll([select Id, CreatedById FROM User limit 10]);

for(SObject record : records) {
    Logger.debug('something, something', record);
}
Logger.saveLog();
```

jongpie:
One screenshot to show Campaign, CampaignMember, Case and Contact records with the related records' names properly set
  
image

jongpie:
And another screenshot to show it working for Lead, Opportunity and Task

image

PR #73, opened 2020-10-21
System mode field is now set based on the current quiddity

### LINKED ISSUE #71: Use Quiddity values for Log__c.SystemMode__c
When this field was first implemented, it was a best-guess approach, and Salesforce did not have the concept of quiddity yet. But it now makes sense to use the current transaction's quiddity enum as the value - the picklist values should be updated to match the list of possible quiddity enums.

jongpie:
Closes #71 with the following changes
- Added `LogEntryEvent__e.SystemMode__c` - this new field, and existing field `Log__c.SystemMode__c`, are now both populated by `Logger.getCurrentQuiddity().name()`
- Updated picklist values in  `Log__c.SystemMode__c` to match the list of [quiddity enum values](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_enum_System_Quiddity.htm#topic-title)
- Eliminated several old fields - using `SystemMode__c` with quiddity seems more robust & easier to maintain
  - `SystemModeIsApexRest__c`
  - `SystemModeIsBatch__c`
  - `SystemModeIsFuture__c`
  - `SystemModeIsQueueable__c`
  - `SystemModeIsScheduledJob__c`
  - `SystemModeIsScheduled__c`


jamessimone:
This was a really good one!

PR #83, opened 2021-02-07
Deprecate most of Aura Log JSON component

### LINKED ISSUE #54: Convert "View JSON" aura component to LWC
We might still need an aura component as a wrapper (since it's a quick action), but the current component should be converted to LWC

jamessimone:
Fixes #54 - Aura wrapper still required since I either misread the release notes for Spring 21, or they have since been updated; LWC quick actions will only be in pilot following Spring 21, not GA.


jongpie:
@jamessimone I deployed the changes to a sandbox - looks good overall, but it seems like the 'copy to clipboard' button isn't working for some reason. I see the JSON being added in console.log & the button visually changes, but nothing is copied to my clipboard. Any thoughts?

jamessimone:
@jongpie thats weird, it was working for me. Let me check a few things!

jongpie:
@jamessimone in case it's a browser-specific issue, I only tested it on Firefox

jamessimone:
@jongpie it very well could be; I also unsuccessfully tested the version I had up here on Firefox this morning. It also could have been carelessness on my part; when I used the _original_ way you were copying data from the Aura component, the stuff that got copied included some new info in LWC specific to the element being appended to the document, which wasn't ideal. It could be that in testing the updates, however, that I hit upon a version that was working, never cleared by clipboard, and then thought something was still working when it wasn't. In any case, I took some inspiration from the method you were originally using (tweaked slightly after reading a Stack Exchange post) and managed to get it once more to only copy the value and nothing to do with the hidden element. This should be working now!

jongpie (approved):

PR #103, opened 2021-03-15
Added log fields for "release number" and "release version"

### LINKED ISSUE #102: Add "Patch Version" field to Log__c
It would be great (even as a toggleable option) to include a remote site setting for trust.salesforce.com so that the patch version the org is currently on within the Log object itself.

The format for the callout is: `https://api.status.salesforce.com/v1/instances/{sfInstanceHere}/status`

jongpie:
Closes #102 with a couple of changes:

- Added 2 new fields to `Log__c`: `ApiReleaseNumber__c` and `ApiReleaseVersion__c`
- `LogEntryEventHandler` now has a private future method that's used to make a callout to `https://api.status.salesforce.com/v1/instances/{sfInstanceHere}/status`
  - To try to minimize the number of callouts made, `LogEntryEventHandler` first tries to query for recent `Log__c` records with the 2 new fields already populated. If a match is found, the existing log's values are copied; if no match is found, the callout is made to the status API.
  - Currently, I have not included a way to enable/disable the callouts. Since the callout happens in a `@future` method, if the remote site setting is disabled, an exception will occur in the future method but it will not impact the log from being saved. I may eventually include a way to prevent the callouts from executing, but I'm skipping that for now.
- Updated page layout and flexipage to show the 2 new fields

image

### DIFF
--- a/nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
+++ b/nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
@@ -10,6 +10,9 @@
 public without sharing class LogEntryEventHandler {
     private static final Log__c LOG = new Log__c();
 
+    @TestVisible
+    private static Boolean shouldCallStatusApi = Test.isRunningTest() == false;
+
     private List<LogEntry__c> logEntries;
     private Map<LogEntry__c, List<String>> logEntryToTopics;
     private Set<String> topicNames;
@@ -100,8 +103,7 @@ public without sharing class LogEntryEventHandler {
         upsert LOG TransactionId__c;
 
         // If no recent logs have the details, and there is not another instance of the job in progress, then start a new one
-        // TODO probably need to remove the check for Test.isRunningTest() == false/add callout mocks to unit tests
-        if (Test.isRunningTest() == false && recentLogWithDetails == null == true && getCountOfOpenJobs() == 0) {
+        if (shouldCallStatusApi == true && recentLogWithDetails == null && getCountOfOpenJobs() == 0) {
             setStatusApiDetails();
         }
     }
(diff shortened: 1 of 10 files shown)


nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
```diff
@@ -209,4 +225,86 @@ public without sharing class LogEntryEventHandler {
...
+
+            logsToUpdate.add(log);
+        }
+        System.debug('logsToUpdate==' + logsToUpdate);
+        update logsToUpdate;
+    }
+
+    // Private class for handling the response from api.status.salesforce.com
+    @TestVisible
+    private class StatusApiResponse {
```
jamessimone, line 302: nice, no need to store _too_ much of that crazy JSON

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -371,4 +399,78 @@ private class LogEntryEventHandler_Tests {
...
+        LogEntryEvent__e logEntryEvent = new LogEntryEvent__e(Message__c = 'my message', Timestamp__c = System.now(), TransactionId__c = transactionId);
+
+
+        Database.SaveResult saveResult;
+
+        Test.startTest();
+
+        saveResult = EventBus.publish(logEntryEvent);
+        // Normally, you don't have to call Test.getEventBus().deliver() if you're also using Test.stopTest()
+        // But, in this case, there are 3 transactions happening (original test, async platform event, and async future method)
```
jamessimone, line 418: thank you for including this note - super helpful

jamessimone (approved):
Looks good to me!

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jongpie: Yeah, I wasted several hours on this trying to get tests to pass - but as soon as I added the calls for `Test.getEventBus().deliver()`, everything finally worked. I thought an extra note would be useful in avoiding this headache again.

nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
jongpie: Yeah, and we can always add new properties to the class if we want to grab anything else from the API.

PR #105, opened 2021-03-16
Added support for logging Flow fault messages

### LINKED ISSUE #98: Add better support for logging Flow fault exceptions
Currently, the `FlowLogEntry` and `FlowRecordLogEntry` classes only have a `message` field - there's not a way to log both a message and an exception message (which _can_ be done in Apex). There needs to either be new optional properties added to the existing classes, or add new classes specifically for logging Flow fault exceptions.

jongpie:
Closes #98 by adding new properties & updated logic in `FlowLogEntry` and `FlowRecordLogEntry`

- Both classes now have an optional string property `faultMessage`
- When the property `faultMessage` is set, the entry's logging level defaults to 'ERROR' (new behavior). When it's null, the default is still 'DEBUG' (existing behavior).
- When the property `faultMessage` is set, the field `LogEntry__c.ExceptionType__c` is automatically set to `Flow.FaultError` and the fault message is stored in the field `LogEntry__c.ExceptionMessage__c` - effectively, the Flow fault error is logged the same way an `Exception` is logged in the Apex methods

### DIFF
--- a/nebula-logger/main/logger-engine/classes/FlowLogEntry.cls
+++ b/nebula-logger/main/logger-engine/classes/FlowLogEntry.cls
@@ -25,6 +25,12 @@ global inherited sharing class FlowLogEntry {
     global String message;
 
     /**
+     * @description Optionally log a Flow fault error message
+     */
+    @InvocableVariable(required=false label='(Optional) Fault Error Message')
+    global String faultMessage;
+
+    /**
      * @description Optionally choose to save any pending log entries.
      */
     @InvocableVariable(required=false label='(Optional) Save Log')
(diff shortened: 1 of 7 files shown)


nebula-logger/main/logger-engine/classes/FlowLogEntry.cls
```diff
@@ -59,6 +65,15 @@ global inherited sharing class FlowLogEntry {
     global static List<String> addFlowEntries(List<FlowLogEntry> flowLogEntries) {
         Boolean saveLog = false;
         for (FlowLogEntry flowLogEntry : flowLogEntries) {
+            // Set the logging level if it's blank
```
jamessimone, line 68: It almost feels like this method should be put into the `LogEntryEventBuilder` and exposed as a public method so that  `FlowRecordLogEntry` and this can share the logic without it being maintained in both places.

jongpie: I definitely agree, I don't like all of the code duplication. But I'll handle that change later - right now, everything is internal code, so I want to look into it a bit more before creating any new `public` methods.

PR #107, opened 2021-03-17
Added Apex managed sharing for Log__c

### LINKED ISSUE #81: Add configurable Apex sharing rules to share records with Log__c.LoggedBy__c
Currently, the project provides 'view all' and 'modify all' access to `Log__c` via the permission sets LoggerLogViewer and LoggerLogAdmin (respectively) - this works for most situations where system admins and power users are the only users monitoring logs. But in some situations, access is also extended to end-users as well; ideally, in the situations, end-users should be able to view their own logs (based on `Log__c.LoggedBy__c`). To support these types of project implementations:

- The project should have a configurable way to enable Apex sharing of `Log__c` records with the logging user. When enabled, the relevant records for `Log__Share` should be generated/maintained after-insert and after-update of `Log__c`
  - A new checkbox field `Log__c.ShareRecordWithLoggedBy__c` (or similar name) should be evaluated by `LogHandler` to handle creating/deleting Apex sharing records
  - `LoggerSettings__c` should have a checkbox field to enable/disable Apex sharing rules (`LoggerSettings__c.EnableShareRecordWithLoggedBy__c` or something similar)
  - `LoggerSettings__c` should have a checkbox field to control the default value of `Log__c.ShareRecordWithLoggedBy__c` allowing it to be configured at the org, profile & user levels (`LoggerSettings__c.DefaultShareRecordWithLoggedBy__c` or something similar)

- The `Log__c` object should use the sharing option 'Grant Access Using Hierarchies' - this will allow managers to view the logs of their team members (in addition to viewing their own logs)
- A batch Apex job might also be needed to handle mass recalculating of Apex sharing for all `Log__c` records
- Bonus points: once Spring '21 is GA, make sure [that manual sharing](https://releasenotes.docs.salesforce.com/en-us/spring21/release-notes/rn_forcecom_manual_sharing_lex.htm) is properly configured for `Log__c` records.

jongpie:
Closes #81 by adding Apex managed sharing for `Log__c`

- On insert, `LogHandler` creates `Log__Share` records for the user specified in `Log__c.LoggedBy__c` with the reason 'Log Created By User'
- New `LoggerSettings__s` field `SharingAccessLevel__c` controls the user's access level (read, edit or none)

**Updated View of Logger Settings**

image

**Example of Sharing Details (in Classic UI)**

image


jongpie:
Adding an updated screenshot for future reference since I ended up (slightly) renaming the custom setting field

image

PR #112, opened 2021-03-19
Added the ability to opt out of status API callout

### LINKED ISSUE #111: Add opt-out for release version/release number callout
Not all orgs are reported on the status.salesforce.com API - being able to gate this callout makes sense for people running private instances of Salesforce. There's no harm in the API call itself; the fields would just be null, but no need to burn the call for an org when we know it will always fail.

jongpie:
Fixes #111 by adding a new field, `LoggerSettings__c.EnableStatusApiCallout__c` - I tried several other options first

- Querying remote site settings - not supported in Apex
- Calling the tooling API to query for remote site settings - feasible, but the callout counts towards the org's daily API limit, and still requires a `@future` method to run, so there's not really any benefit
- Tried to mimic @jamessimone 's screen-scraping in [Apex space storage repo](https://github.com/jamessimone/apexStorageSpace) - the screen-scraping worked really well, but since it involves a `PageReference`, it can't be called from a trigger. The only workaround would be to run it in a `@future` method, but again, that defeats part of the goal of #111

So for now, the new custom setting field is the best option - the code only checks the setting at the org level, using `LoggerSettings__c.getOrgDefaults()`


jamessimone:
Looks good to me!

PR #114, opened 2021-03-21
Added support for lists of database result & sobject parameters

### LINKED ISSUE #101: Add support for bulk Database result methods
We recently added options (#59)  to log `Database.SaveResult` and the other DML-based result classes, but these classes are most typically returned to the client in bulk (ie, `List<Database.SaveResult>`). For the times where the user just wants to create a log of all the results, we should expose bulk methods as well to better assist so that they don't have to iterate through each item in the collection manually.

### LINKED ISSUE #115: Setting Database Results
When I add logs that take Database Results as a parameter and User executing the code doesn't have logging enabled (for example, in a test class that don't have LoggerSettings__c initialized), code is failing with exception.

```
System.NullPointerException: Attempt to de-reference a null object
Class.LogEntryEventBuilder.setDatabaseResult: line 187, column 1
Class.Logger.warn: line 969, column 1
```

The reason for this it that LogEntryEventBuilder.setDatabaseResult methods are not checking, if logging is enabled.
The missing code is:

```
if (this.shouldSave == false) {
    return this;
}
```

Without this code, method is trying to reference this.logEntryEvent.DatabaseResultJson__c, which is not initialized - LogEntryEventBuilder constructor also have condition, that it doesn't initialise logEntryEvent when shouldSave variable is set to False

jongpie:
Closes #101 by adding several new methods

- `LogEntryEventBuilder` now has overloads for `setDatabaseResult()` and `setRecord()` that accept lists as parameters
- `Logger` now has an obnoxious number of overloads for each logging level that accept lists as parameters
- Added fields `LogEntry__c.DatabaseResultCollectionType__c` and `LogEntry__c.RecordCollectionType__c` to track if the parameters are `Single` (single record) or `List` (list of records)

Also fixes #115 by adding missing guard clauses in `LogEntryEventBuilder`


nebula-logger/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -141,93 +141,263 @@ global with sharing class LogEntryEventBuilder {
     // DML-result builder methods. All* of the Result classes behave the same (*except Upsert)
     // ...but Salesforce didn't bother to use an interface for the classes, so instead - overloads!
     /**
-     * @description Sets the log entry event's DML result fields
+     * @description Sets the log entry event's database operation result fields
      * @param  deleteResult The instance of `Database.DeleteResult` to log
      * @return              The same instance of `LogEntryEventBuilder`, useful for chaining methods
      */
     global LogEntryEventBuilder setDatabaseResult(Database.DeleteResult deleteResult) {
+        if (this.shouldSave == false) {
```
jamessimone, line 149: seems like these two `if` statements in each method can be combined!

nebula-logger/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -423,6 +599,10 @@ global with sharing class LogEntryEventBuilder {
     }
 
     private String getSObjectClassification(Schema.SObjectType sobjectType) {
+        if (sobjectType == null) {
```
jamessimone, line 602: safe navigation here can cut down on the complexity here and for the next addition (~ line 639) by reducing the need for these if statements!

nebula-logger/tests/logger-engine/classes/LogEntryEventBuilder_Tests.cls
```diff
@@ -374,10 +383,246 @@ private class LogEntryEventBuilder_Tests {
...
+
+        Test.startTest();
+        builder.setDatabaseResult(upsertResults);
+        Test.stopTest();
+
+        LogEntryEvent__e logEntryEvent = builder.getLogEntryEvent();
+
+        System.assertEquals('List', logEntryEvent.DatabaseResultCollectionType__c);
+        System.assertEquals(JSON.serializePretty(upsertResults), logEntryEvent.DatabaseResultJson__c);
+        System.assertEquals(Database.UpsertResult.class.getName(), logEntryEvent.DatabaseResultType__c);
```
jamessimone, line 597: mmm, this is an interesting one. I guess that we just say, hey, the JSON is there in the event that you need to examine the results of `isCreated()` for each result

jamessimone:
Just a few comments - I think some opportunities to cut down on the lines of code with a few `if` statements that can be condensed / gotten rid of altogether in conjunction with safe navigation - otherwise, looks good to me!

nebula-logger/main/logger-engine/classes/LogEntryEventBuilder.cls
jongpie: I thought about using safe navigation, but using a guard clause at the top cuts out the rest of the methods logic from running. If I switch to using safe navigation (and remove the guard clause), then the full method executes, which eats up some CPU usage.

nebula-logger/main/logger-engine/classes/LogEntryEventBuilder.cls
jongpie: Yeah, good call, I'll consolidate the guard clauses

PR #133, opened 2021-04-19
Capture log events at the millisecond level

arvindnarsimhan:
Added TimestampInMilliSecond__c field to capture log events at the millisecond level.


nebula-logger/main/logger-engine/objects/LogEntryEvent__e/fields/TimestampInMilliSecond__c.field-meta.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<CustomField xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>TimestampInMilliSecond__c</fullName>
```
jongpie, line 3: Minor detail, but for consistency, can you make the 'S' in 'MilliSecond__c' lowercase (i.e., `TimestampInMillisecond__c`)?

arvindnarsimhan: @jongpie Appreciate your detail for consistency. Fixed the field name. Let me know.

nebula-logger/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -65,6 +65,7 @@ global with sharing class LogEntryEventBuilder {
             LoggingLevelOrdinal__c = loggingLevel.ORDINAL(),
             OriginType__c = ORIGIN_TYPE_APEX,
             Timestamp__c = System.now(),
+            TimestampInMillisecond__c = DateTime.now().getTime(),
```
jongpie, line 68: Hey @arvindnarsimhan - one last thought on the field name. Since this is now using the `getTime()` method instead of `milliseconds()`, I think a field name like `EpochTimestamp__c` would be more accurate/descriptive than `TimestampInMillisecond__c`. Let me know what you think.

arvindnarsimhan: @jongpie Let me rename it to EpochTimestamp__c

arvindnarsimhan:
@jongpie Renamed the field to EpochTimestamp as suggested. Don't hesitate if you have more changes.

jongpie:
Thanks @arvindnarsimhan for renaming the field! The label on `nebula-logger/main/logger-engine/objects/LogEntryEvent__e/fields/EpochTimestamp__c.field-meta.xml` is still set to 'TimestampInMillisecond` - can you update the label to 'Epoch Timestamp' so it matches the field API name?

And it looks like there are still 2 versions of the field `TimestampInMilliSecond__c` included in this PR. Can you run these 2 commands & push the changes?
```shell
git rm nebula-logger/main/logger-engine/objects/LogEntryEvent__e/fields/TimestampInMillisecond__c.field-meta.xml
git rm nebula-logger/main/logger-engine/objects/LogEntryEvent__e/fields/TimestampInMilliSecond__c.field-meta.xml
```

I think this is the last set of changes needed, and then we can merge this!

arvindnarsimhan:
@jongpie - My bad for not fixing the label name. Thanks for catching it. I have also cleaned up the files. Let me know.

PR #135, opened 2021-04-20
Fix System.QueryException: unexpected token: '('

daveerickson:
Fix for unexpected difference in how SOQL and Dynamic SOQL bind to Network.


jongpie:
@daveerickson thanks for working on this! I tested it in 2 scratch orgs (1 with communities enabled, and 1 without), and it looks like everything's working now. I'll merge your changes.

PR #142, opened 2021-05-03
Release v4.4.0

jongpie:
- Added new field `LogEntry__c.EpochTimestamp__c` (set based on `LogEntryEvent__e.EpochTimestamp__c`)
- Updated list of preview and non-preview instances for Summer '21 release
* Added new public methods `Logger.getUserSettings()` and `Logger.getUserSettings(User loggingUser)`
- Update handler classes to allow tests to mock trigger-based variables


jamessimone:
I went to look for a place where I could comment on that View Documentation button upgrade, and this is where I got to 😇.

PR #147, opened 2021-05-18
Fix DML chunking error potential

### LINKED ISSUE #146: LogBatchPurger fails with error message Cannot have more than 10 chunks in a single operation. Please rearrange the data to reduce chunking.
The error :"Cannot have more than 10 chunks in a single operation. Please rearrange the data to reduce chunking." is due to the arrangement of the recordsToDelete list in LogBatchPurger on line 50. 

Records are being added like so: 

> `for (Log__c log : scope) {
        recordsToDelete.addAll(log.LogEntries__r);
        recordsToDelete.add(log);
}`

This means you can end up with a list like this: 
>Log__c, LogEntry__c, LogEntry__c, Log__c, LogEntry__c, Log__c, LogEntry__c, Log__c, LogEntry__c, Log__c, LogEntry__c.

Apex will only chunk up to 10 times, this means your generic Sobject list needs to be sorted by Sobject type(to allow a generic deletion) like so:
>Log__c, Log__c, Log__c, Log__c, LogEntry__c, LogEntry__c, LogEntry__c, LogEntry__c.......

The above will chunk 2 times, one for the Log__c, one for the LogEntry__c.

A simple fix(but probably inefficient in a batch) is to call recordsToDelete.sort();

Probably a nicer way to do it is to have each SObject in it's own deletion list and the add them together once you've collected all the data. See below for example:
> `List<Log__c> logs = [SELECT Id, (SELECT Id FROM LogEntries__r) FROM Log__c LIMIT 500];

>List<Log__c> logsToDelete = new List<Log__c>();
>List<LogEntry__c> logsEntriesToDelete = new List<LogEntry__c>();

>for (Log__c log : logs) {
>    logsToDelete.add(log);
>    logsEntriesToDelete.addAll(log.LogEntries__r);
>}

>List<SObject> allRecords = new List<SObject>();
>allRecords.addAll(logsEntriesToDelete);
>allRecords.addAll(logsToDelete);
>delete allRecords;`

I did try implementing this but it wasn't playing nice with your test classes. If I get time I will look again. 

Thanks for all the work on this amazing Logger!

jamessimone:
Fixes #146

### DIFF
--- a/nebula-logger/main/log-management/classes/LogBatchPurger.cls
+++ b/nebula-logger/main/log-management/classes/LogBatchPurger.cls
@@ -47,11 +47,20 @@ global with sharing class LogBatchPurger implements Database.Batchable<SObject>,
                 Logger.info(new LogMessage('Starting deletion of {0} records', scope.size()));
             }
 
-            List<SObject> recordsToDelete = new List<SObject>();
+            List<Log__c> logsToDelete = new List<Log__c>();
+            List<LogEntry__c> logEntriesToDelete = new List<LogEntry__c>();
             for (Log__c log : scope) {
-                recordsToDelete.addAll(log.LogEntries__r);
-                recordsToDelete.add(log);
+                logEntriesToDelete.addAll(log.LogEntries__r);
+                logsToDelete.add(log);
             }
+
+            List<SObject> recordsToDelete = new List<SObject>();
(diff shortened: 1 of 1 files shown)


jongpie (approved):
@jamessimone looks good to me! Merge away!

PR #163, opened 2021-06-04
Added new save method 'SYNCHRONOUS_DML'

jongpie:
This skips publishing platform events and instead immediately creates Log__c and LogEntry__c records

### DIFF
--- a/nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
+++ b/nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
@@ -37,13 +37,11 @@ public without sharing class LogEntryEventHandler {
      * @description Runs the trigger handler's logic for the `LogEntryEvent__e` platform event object
      */
     public void execute() {
-        if (this.logEntryEvents.isEmpty()) {
-            return;
+        if (this.logEntryEvents.isEmpty() == false) {
+            this.upsertLogs();
+            this.insertLogEntries();
+            this.insertTopics();
         }
-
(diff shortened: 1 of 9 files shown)


nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
```diff
@@ -36,13 +37,13 @@ public without sharing class LogEntryEventHandler {
      * @description Runs the trigger handler's logic for the `LogEntryEvent__e` platform event object
      */
     public void execute() {
-        switch on this.triggerOperationType {
-            when AFTER_INSERT {
-                this.upsertLogs();
-                this.insertLogEntries();
-                this.insertTopics();
-            }
+        if (this.logEntryEvents.isEmpty()) {
```
jamessimone, line 40: something that I think helps with the readability/avoidance of an early return is inverting the conditional - if they list is _not_ empty, upsert logs, insert log entries, etc ...

jamessimone:
@jongpie one minor comment, this looks great!

nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
jongpie: You've convinced me! I've flipped it around so it calls the private methods if the list is not empty.

jongpie:
@jamessimone thanks for reviewing!

jamessimone (approved):

PR #165, opened 2021-06-04
Added plugin framework + re-added Slack integration

### LINKED ISSUE #164: Add the ability to extend the SObject handler classes
There should be a way (at least in Apex - potentially Flow as well) to run additional logic in the `LogHandler`, `LogEntryHandler` and `LogEntryEventHandler` classes without having to modify these classes directly.

- Each handler class should manage running any extra logic for its SObject Type
- Developers should be able to write their own Apex classes to further manipulate the logging records in `Log__c` and `LogEntry__c`, `LogEntryEvent__e`

### LINKED ISSUE #169: Re-add Slack integration
In [an old beta release of Logger](https://github.com/jongpie/NebulaLogger/releases/tag/v0.10.0-beta), there was built-in support for sending alerts to Slack for certain logs/log entries - I ended up [removing it](https://github.com/jongpie/NebulaLogger/releases/tag/v2.1.0-beta) to focus on the core logging functionality, but it'd be great to have an add-on package that can be installed on top of Logger in orgs that want to integrate with Slack.

**Design**

- The entire integration should be 100% optional - not every org uses Slack, so no extra metadata for this integration should be included in the [nebula-logger](https://github.com/jongpie/NebulaLogger/tree/main/nebula-logger) directory
    - TODO: need to decide if this will (dynamically?) work for both versions of the package (unlocked & managed w/ namespace), but to start, it'll probably be just for the unlocked package (no namespace)
- There should be a way to control which logs/log entries are pushed to Slack - the old integration used a checkbox field on the records, that may still be the best way to handle it
- There should be a declarative way to control the message string sent to Slack - the base message could be stored as a custom label or in a CMDT object
- There should be a way to declaratively set/update the Slack details (API key, channel, etc.)

jongpie:
Closes #164 and closes #169 by introducing a few new features for the trigger handler classes. This is currently only available in the unlocked package - there are some technical concerns & constraints with bringing this functionality into the managed package.

### New Logger plugin framework (beta)
- Added new abstract class `LoggerSObjectHandler` - the handlers for `Log__c`, `LogEntry__c` and `LogEntryEvent__e` now extend this class
- Added new CMDT object `LoggerSObjectHandler__mdt` to control the behavior of the handler classes
    - `IsEnabled__c` - controls if the handler class runs at all
    - `SObjectType__c` - entity definition-relationship field to indicate which SObject is processed (`Log__c`, `LogEntry__c` and `LogEntryEvent__e`)
- Added new CMDT object `LoggerSObjectHandlerPlugin__mdt` to define & control any plugins that the handler class should execute
    - `SObjectHandler__c` - a metadata lookup field, used to relate a plugin to an instance of `LoggerSObjectHandler__mdt`. This controls when your plugin will execute
    - `PluginType__c` - the type of plugin to execute. Possible values are 'Apex' and 'Flow'
    - `PluginApiName__c` - the name of an Apex class or Flow to automatically execute
        - Type == 'Apex' - the name of an Apex class that implements the interface `LoggerSObjectPostProcessor`
        - Type == 'Flow' - the name of a Flow with the input parameters `triggerOperationType`, `records` and `oldRecords`
- Added new CMDT object `LoggerSObjectHandlerPluginParameter__mdt` for plugins to store key-value pair configurations

### New Slack plugin for Logger (beta)
- Re-added the old Slack integration using the new plugin framework. Formerly a batch job, the new Slack plugin uses a `queueable` job for asynchronously sending Slack notifications for logs that meet a threshold logging level


jamessimone:
Nice. Looks like this one is shaping up quickly!

jongpie:
@jamessimone I made a few small-but-important changes, I'd love to know your thoughts whenever you have a chance:

- I changed `LoggerHandler` to be a `public abstract` class that all 3 handler classes now extend (to allow sharing code between the handlers, like a mini trigger framework)
- The interface `LoggerHandler.PostProcessor` has been removed from `LoggerHandler` and it's now a standalone `global interface` called `LoggerPostProcessor`

nebula-logger/main/log-management/classes/LogEntryHandler.cls
```diff
@@ -39,6 +42,9 @@ public without sharing class LogEntryHandler {
                 this.setCheckboxFields();
             }
         }
+
+        // Run any post-processors
+        super.executePostProcessors(this.logEntries, this.oldLogEntriesById);
```
jamessimone, line 47: I would avoid using `super` here because you're not overriding the `executePostProcessors` method - `this.executePostProcessors(...)` will suffice!

nebula-logger/main/log-management/classes/LogHandler.cls
```diff
@@ -58,6 +58,9 @@ public without sharing class LogHandler {
                 this.shareLogsWithLoggingUsers();
             }
         }
+
+        // Run any post-processors
+        super.executePostProcessors(this.logs, this.oldLogsById);
```
jamessimone, line 63: same here

jamessimone (changes requested):
Two minor comments!

nebula-logger/main/log-management/classes/LogEntryHandler.cls
jongpie: Yeah, I couldn't decide if using `super` or `this` was more intuitive - I'll switch to `this` instead!

nebula-logger/main/log-management/classes/LogHandler.cls
```diff
@@ -38,10 +38,22 @@ public without sharing class LogHandler {
...
+     */
+    public override SObjectType getSObjectType() {
+        return Schema.Log__c.SObjectType;
+    }
+
     /**
      * @description Runs the trigger handler's logic for the `Log__c` custom object
      */
     public void execute() {
+        if (this.configuration.IsEnabled__c == false) {
```
jamessimone, line 53: It seems like these checks could be consolidated in the conditionals you have at the beginning of the base class’s “executePostProcessorsFor...” methods so that the subclasses don’t always have to remember to do that check. That probably allows the configuration object to not have to be protected visibility, too

jongpie: The problem is that this guard clause is intended to disable the entire handler's execution, not just the post-processors - I've been meaning to add a way to allow handlers to be temporarily disabled (since you can't disable a trigger in a prod org without deploying), but I think it requires keeping these guard clauses within each handler's `execute()` method

derroman:
We want to see that! :-D

jamessimone:
@derroman it’s nearly ready!

derroman:
need a helping hand? ;-)

jongpie:
@derroman it's ready! I've merged & released these changes - let us know if you have any feedback!

derroman:
_[cancels every meeting of today]_ 😂
Great! Looking forward to play around with!

PR #173, opened 2021-06-27
New custom tagging objects + new tagging features

### LINKED ISSUE #90: Creating TopicAssignment records silently fails in a managed package context
The `Logger` and `LogEntryEventBuilder` methods for using Salesforce's Topics are not currently available in the managed package due to an issue with the functionality. New `Topic` records were successfully created, but the related `TopicAssignment` records for `Log__c` and LogEntry__c` were never inserted. 

This functionality all works both when running in an org without a namespace (unpackaged code), as well as in a namespaced scratch org (using the same `Nebula` namespace, but silently fails in orgs that have installed the managed package.

Possible theories
- Some kind of permission issue with managed package + automated process user
- Maybe topic assignments should be created async (but still would not explain why this functionality works fine when running as unpackaged code/no namespace)

### LINKED ISSUE #106: Tagging on the Log__c record
I'm trying to debug a specific apex class and is there a way to easily tag a specific Log__c so that I'll know what it is for?

I noticed the following in the documentation: Assign Topics (Tagging/Labeling System)

I'm actually using the Managed Package to limit the maintenance on client orgs. I'm interested to hear what specific issue is preventing it from being included in the managed package?

### LINKED ISSUE #116: Add configurable topic assignment based on LogEntry__c fields
In addition to the current functionality for assigning topics, there should also be a way (using CMDT) to dynamically assign additional topics to `Log__c` and `LogEntry__c` based on the value of `LogEntry__c` fields. This will allow additional topics to be assigned without modifying the underlying Apex or Flow automations.

### LINKED ISSUE #154: Topic creation failure with spaces as differentiator
I know this is a little late to the party since you pulled out the public Topic methods for redesign, however, perhaps it can save you or someone else some pain down the road.

I was using Topics (successfully), and since you left the setTopics in the LogEntryEventBuilder I will continue using them as I was. with the Unlocked package.

I stumbled on this when I was doing some refactoring, I removed a space from one of the Topics I was using and then I was not getting any topics on my log entry outputs.

It appears that inserting topics with only spaces as a differentiator in the name causes a DML DUPLICATE_VALUE exception.

Since, the logic to look up the Topics by name returns no results for the 'spaced' version it is considered new and subsequently an exception is thrown when it attempts the insert.

### LINKED ISSUE #166: LogBatchPurger Too many DML rows: 10001
Getting this error for a batch whenever LogBatchPurger runs.

Almost latest version 4.4.5 of the Unlocked Package.

Will try to get some more details.

### LINKED ISSUE #174: Get "INVALID_FIELD, Foreign key external ID: .... not found for field TransactionId__c" when
When using `setParentLogTransactionId(logger.getTransactionId());` to assign a value equivalent to the current id, the DML operation fails...

The receiving method should be adjusted so it only passes the value downstream if the value is different than the current session.

### LINKED ISSUE #176: LogBatchPurger - Apex CPU time limit exceeded
Having issues with the LogBatchPurger, admittedly I'm logging a lot of things.

Looking into the class, the start method is also querying for the Id from LogEntries__r relationship, but it appears to be unused.

Removing this part from the start function seems to resolve this issue for me and as far as I can tell does not affect the process.

Can you advise on anything I'm overlooking and a reason why this would be a bad idea?

Thanks

### LINKED ISSUE #178: Is Epoch Sorting available in Logger Console?
When using Logger Console the Timestamp column for the LogEntry most times is not helpful since it displays limited precision, is there a way to add precision to this, if not is there a way to add the Epoch for sorting?

I have found sometimes that my log entries are not in the correct order and navigating to the record to compare epochs is frustrating and time consuming.

Thanks

image

image

jongpie:
Closes #90, closes #106, closes #116, fixes #166, fixes #154, fixes #174, fixes #176 and closes #178 🎉

### Tagging System Overhaul
One of [the oldest features within Nebula Logger](https://github.com/jongpie/NebulaLogger/releases/tag/v0.9.0-beta) is tagging your log entries by using the `Topic` and `TopicAssignment` Chatter objects. The overall data model & functionality of Topics is great for a tagging/labeling system, but it's proven to have some issues over the last few years, including:

- Not everyone uses Chatter
- Not everyone that does use Chatter wants logging tags (Topics) to be visible to all users - but `Topic` records are visible to everyone in Chatter
- Reporting & dashboard functionality for Topics is [only partially implemented](https://trailblazer.salesforce.com/ideaView?id=08730000000l12wAAA)
- #90 - Even though the same code works in the unlocked package, I could not get the functionality to fully work when running within a namespace context. This prevents the managed package from being able to use the Topics-based tagging system at all

This PR introduces some new objects & logging options to try to solve a lot of these issues + introduce some more robust tagging functionality
- 2 "tagging modes" are now built into Nebula Logger
  1. Using Salesforce Topics: this old functionality is still available in the unlocked package (but still not available in the managed package. However, as of release v4.6.0, it will no longer be the default. After upgrading to v4.6.0, you can revert to using Topics by updating the custom metadata record "LogEntryEventHandler - Use Topics for Tags" within the CMDT object `LoggerSObjectHandlerParameter__mdt` - simply change the `Value__c` field from `false` to `true` and save.
  2. Using Logger's Custom Tagging objects: available in both the unlocked package and managed package, this mode will be the default as of release v4.6.0. This mode uses 2 additional custom objects, `LoggerTag__c` and `LogEntryTag__c` in place of `Topic` and `TopicAssignment` (respectively)
- Tags can be added in 3 different ways
  1. **In Apex**: developers can use 2 new methods in `LogEntryBuilder` to add tags
      ```java
      // Use addTag(String tagName) for adding 1 tag at a time
      Logger.debug('my log message').addTag('some tag').addTag('another tag');

      // Use addTags(List<String> tagNames) for adding a list of tags in 1 method call
      List<String> myTags = new List<String>{'some tag', 'another tag'};
      Logger.debug('my log message').addTags(myTags);
      ```
  2.  **In Flow**: flow builders can set the property `List<String> tags` within the Flow classes `FlowLogEntry`, `FlowRecordLogEntry` and `FlowCollectionLogEntry`
  3. **Using the CMDT object `LogEntryTagRule__mdt`**: admins can configure tagging rules to append additional tags. These tags are added on top of any tags added via Apex and/or Flow. Rules can be set up by configuring a rule with these fields set:
      1. Logger SObject: currently, only the "Log Entry" object (`LogEntry__c`) is supported.
      2. Field: the SObject's field that should be evaluated - for example, `LogEntry__c.Message__c`. Only 1 field can be selected per rule, but multiple rules can use the same field.
      3. Comparison Type: the type of operation you want to use to compare the field's value. Currently supported options are: `CONTAINS`, `EQUALS`, `MATCHES_REGEX`, and `STARTS_WITH`
      4. Comparison Value: the comparison value that should be used for the selected field operation
      5. Tags: a list of tag names that should be dynamically applied to any matching `LogEntry__c` records. When specifying multiple tags, put each tag on a separate line within the Tags field
      6. Is Enabled: only enabled rules are used by Logger - this is a handy way to easily enable/disable a particular rule without having to entirely delete it


nebula-logger/main/log-management/flexipages/LoggerConsoleUtilityBar.flexipage-meta.xml
```diff
@@ -1,11 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <FlexiPage xmlns="http://soap.sforce.com/2006/04/metadata">
     <flexiPageRegions>
-        <itemInstances>
```
jamessimone, line 4: Aw yeah

README.md
```diff
@@ -346,6 +346,77 @@ Within Flow (and Process Builder), there are 4 invocable actions that you can us
 
 ---
 
+## Tagging Your Log Entries
+
+Nebula Logger supports dynamically tagging/labeling your `LogEntry__c` records via Apex and Flow. Tags can then be stored using one of the two supported modes (discussed below).
+
+### Adding Tags in Apex
+
+// TODO LogEntryEventBuilder.addTag(String tagName) and addTags(List<String> tagNames)
```
jamessimone, line 355: are these upcoming (post PR) changes?

README.md
```diff
@@ -346,6 +346,77 @@ Within Flow (and Process Builder), there are 4 invocable actions that you can us
...
+                    <li><code>TopicAssignment</code>: a junction object between a <code>Topic</code> record and any (supported) SObject, using a polymorphic <code>EntityId</code> field. See <a href="https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_topicassignment.htm">object reference docs for more details</a></li>
+                </ul>
+            </td>
+        </tr>
+        <tr>
+            <td>Data Visibility</td>
+            <td>
+                <ul>
+                    <li>Access to the <code>LoggerTag__c</code> object can be granted/restricted using standard Salesforce object and record-sharing functionality (OWD, sharing rules, profiles, permission sets, etc). By default, <code>LoggerTag__c</code> OWD is set to 'public read-only' for internal users and 'private' for external users</li>
+                    <li>Since <code>LogEntryTag__c</code> is junction object, access to these records is controlled by a user's access to the related <code>LogEntry__c</code> and <code>LoggerTag__c</code> records</li>
```
jamessimone, line 399: is a* junction object

nebula-logger/main/log-management/classes/LogBatchPurgeScheduler.cls
```diff
@@ -9,13 +9,23 @@
  * @see LogBatchPurger
  */
 global with sharing class LogBatchPurgeScheduler implements System.Schedulable {
+    private Integer batchSize;
```
jamessimone, line 12: if you're not updating this value, I would mark it as `final`

nebula-logger/main/log-management/classes/LogBatchPurger.cls
```diff
@@ -9,101 +9,47 @@
  * @see LogBatchPurgeScheduler
  */
 global with sharing class LogBatchPurger implements Database.Batchable<SObject>, Database.Stateful {
```
jamessimone, line 11: this whole update to `LogBatchPurger` is great

nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
```diff
@@ -218,64 +221,173 @@ public without sharing class LogEntryEventHandler extends LoggerSObjectHandler {
...
+                missingTagNameToId.put((String) tag.get('Name'), (Id) tag.get('Id'));
             }
+        }
+
+        return missingTagNameToId;
+    }
+
+    // Private static methods
+    private static String getLogEntryKey(LogEntry__c logEntry) {
+        // TODO - after upgrading all metadata to v52.0, switch to using the new EventUuid field for platform events
```
jamessimone, line 332: nice

README.md
jongpie: Oops, I overlooked these TODOs - I'm going to include these in this PR, I'll try to finish up the README changes tonight

nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
```diff
@@ -218,64 +221,173 @@ public without sharing class LogEntryEventHandler extends LoggerSObjectHandler {
...
+        }
+        return tagAssignmentRules;
+    }
+
+    private static Boolean ruleCriteriaMet(LogEntry__c logEntry, LogEntryTagRule__mdt rule) {
+        Boolean ruleCriteriaMet = false;
+        String logEntryFieldValue = String.valueOf(logEntry.get(rule.SObjectField__c));
+        String ruleComparisonValue = rule.ComparisonValue__c;
+
+        switch on rule?.ComparisonType__c.toUpperCase() {
```
jamessimone, line 370: is there a test for comparison type being null? I haven't tried switching on null when the type wasn't known ...

jongpie: There's not a test at the moment - it should just get ignored (since it doesn't match one of the specified Strings in the switch statement), but I'll add one just to be sure that it doesn't break anything

nebula-logger/main/logger-engine/classes/FlowLogger.cls
```diff
@@ -42,14 +43,18 @@ public inherited sharing class FlowLogger {
...
-                    this.loggingLevelName = 'DEBUG';
+                    this.loggingLevelName = LoggingLevel.DEBUG.name();
                 }
             }
 
-            this.loggingLevel = Logger.getLoggingLevel(this.loggingLevelName);
-            this.logEntryEventBuilder = Logger.newEntry(this.loggingLevel, this.message).setTopics(this.topics);
+            // In Flow, using comma-separated String is easier than List<String>
+            // but LogEntryEventBuilder.addTags() uses List<String>, so FlowLogger handles the conversion work
+            List<String> parsedTags = this.tagsString?.replace(', ', ',').replace(' ,', ',').split(',');
```
jamessimone, line 54: not sure which would be more effective, but you could use `( ,)|(,)|(, )` to do this in one replace

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -474,18 +215,58 @@ private class LogEntryEventHandler_Tests {
         }
     }
 
+    // TODO - testing topics is tricky. Within the unlocked package, we can test it,
```
jamessimone, line 218: planning to remove this test?

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -474,18 +215,58 @@ private class LogEntryEventHandler_Tests {
...
+    //     for (TopicAssignment topicAssignment : topicAssignments) {
+    //         System.assert(tagsSet.contains(topicAssignment.Topic.Name));
+    //     }
+    // }
+
     @isTest
-    static void it_should_create_topic_and_topicAssignment_records() {
-        String transactionId = '123-456-789-0';
+    static void it_should_create_tag_records_when_enabled() {
+        System.assertEquals(true, LogEntryEventHandler.TAGGING_IS_ENABLED);
```
jamessimone, line 260: I don't always have custom assert messages, but especially since this is at the start of the test I think something about the test having started under the wrong conditions helps to clarify what's happening in the event of a premature failure

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -502,13 +283,214 @@ private class LogEntryEventHandler_Tests {
...
+        System.assertEquals(tags.size(), logEntryTags.size());
+        Set<String> tagsSet = new Set<String>(tags);
+        for (LogEntryTag__c logEntryTag : logEntryTags) {
+            System.assert(tagsSet.contains(logEntryTag.Tag__r.Name));
         }
     }
 
+    @isTest
+    static void it_should_reuse_existing_tag_records() {
+        System.assertEquals(true, LogEntryEventHandler.TAGGING_IS_ENABLED);
```
jamessimone, line 296: same here

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -502,13 +283,214 @@ private class LogEntryEventHandler_Tests {
...
+        Integer countOfTagsWithTagName = [SELECT COUNT() FROM LoggerTag__c WHERE Name = :testTagName];
+        System.assertEquals(1, countOfTagsWithTagName);
+        LogEntryTag__c logEntryTag = [SELECT Id, LogEntry__c, Tag__c, Tag__r.Name FROM LogEntryTag__c WHERE LogEntry__c = :logEntry.Id];
+        System.assertEquals(tag.Id, logEntryTag.Tag__c);
+        System.assertEquals(tag.Name, logEntryTag.Tag__r.Name);
+    }
+
+    @isTest
+    static void it_should_append_tags_for_contains_rule() {
+        System.assertEquals(true, LogEntryEventHandler.TAGGING_IS_ENABLED);
```
jamessimone, line 334: same here

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -691,4 +672,264 @@ private class LogEntryEventHandler_Tests {
...
+            orgEnvironmentType = 'Scratch Org';
+        } else if (org.IsSandbox == true) {
+            orgEnvironmentType = 'Sandbox';
+        } else {
+            orgEnvironmentType = 'Production';
+        }
+
+        System.assertEquals(logEntryEvent.ApiVersion__c, log.ApiVersion__c);
+        System.assertEquals(logEntryEvent.Locale__c, log.Locale__c);
+        //System.assertEquals(logEntryEvent.CreatedById, log.LoggedBy__c);
```
jamessimone, line 842: remove these commented out fields?

nebula-logger/tests/log-management/classes/LogMassDeleteExtension_Tests.cls
```diff
@@ -107,8 +107,19 @@ private class LogMassDeleteExtension_Tests {
         PageReference pageReference = Page.LogMassDelete;
         Test.setCurrentPage(pageReference);
 
-        User chatterUser = [SELECT Id FROM User WHERE Profile.Name = 'Chatter Free User' LIMIT 1];
-        System.runAs(chatterUser) {
+        Profile standardUserProfile = [SELECT Id FROM Profile WHERE Name = 'Standard User' LIMIT 1];
```
jamessimone, line 110: thoughts on abstracting this out?

nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
jongpie: Actually, I'm going to skip adding a test for null - the field `ComparisonType__c` is required, so null wouldn't actually happen.

Nevertheless, I wrote this small `switch` statement to show how nulls are handled - the null falls into the `else` block. If there weren't an `else` section, then the null value would just be ignored altogether by the `switch` statement
```java
String someString = null;
switch on someString {
    when 'something' {
        System.debug('someString == something');
    }
    when else {
        System.debug('here we are in the ELSE block, someString==' + someString);
    }
}
```

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jongpie: I'm going to leave it for now, since it is a functioning test, it just can't be tested in all orgs. I'm planning to revisit it after this PR - it'll probably get moved to the `extra-tests` folder

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jongpie: Yeah, that's completely fair - I need to be better about adding assert messages. I don't think they're needed for every assert, but there are definitely some asserts in this repo that would benefit from a custom message.

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jongpie: I think I meant to solve this in an older PR - the issue is that `logEntryEvent.CreatedById` is null in this context (it's populated in the platform-event trigger), and there's not a way to re-query the platform event like you would a "normal" SObject.

But it turns out, I can just use `UserInfo.getUserId()` instead, so I'll uncomment + fix these 2 fields

nebula-logger/main/log-management/classes/LogEntryEventHandler.cls
jamessimone: Ah yeah, I was aware of the behavior of null in the when blocks - I guess I was just surprised that the compiler actually knows that safe navigation on a case statement that results in null is fine (basically it seems it must infer the type from the field); in other words, I don’t think that `switch on null` is currently valid, but clearly there was some thought put into how safe navigation works there

nebula-logger/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jongpie: Custom assert messages have been added! 🎆

nebula-logger/tests/log-management/classes/LogMassDeleteExtension_Tests.cls
jongpie: It's been moved to a new class `LoggerTestUtils` - I should have created this class a long time ago, there's some other duplication in existing test classes that could/should be consolidated. I'll move more test methods to this class in a future PR.

PR #185, opened 2021-08-20
New 'Filtered Log Entry Events' plugin

### LINKED ISSUE #183: Selective event listening
Hi @jongpie 

The Nebula Logger works very well, and we get great benefit from it, listening remotely using a CometD client.

However, we have encountered an issue and are looking for a solution:

We have an enormous amount of logs in our system. 

 
Problem is, we reach the Salesforce Organization daily limit, and get blocked from the rest of the events, on a daily basis. 

**We were thinking that the ability to control which logs we listen to, would help us prevent this:** 
Regularly listen to critical log events, and in case needed - listen to informative events to track down errors, as salesforce keeps the events in the event bus.

We did take advantage of the LoggingLevel, separating logs into different levels, mainly critical logs(Errors, exceptions, etc.) and informative logs (callouts, queries, etc.)

Selective listening, of course, cannot be achieved by changing log levels, can it? we still need all logs to be created, just listen to some of them, and have the others at hand in case they are needed. 

Can you think of, or are you working on any solution for this? 

We did come up with the idea to create separate events for each log level (i.e. LogEntryEventError__e, LogEntryEventDebug__e etc.), that will be created along with the creation of LogEntryEvent__e. (which means double the events created). 

An expansion of that idea is just to add 2-3 more event types (e.g. LogEntryEventHigh__e, LogEntryEventLow__e etc. ) to represent high/low levels of events, and allow the user to decide which logging levels are mapped as High/Low levels of logs, via the LoggerSettings custom setting.

This way, for example, a user could assign some logging levels (e.g. ERROR, WARN, DEBUG) as HIGH, allowing listening only to these logs regularly, while other logging levels (INFO, FINE etc.) could be assigned to LOW level, and listened to on demand. 

Is that something that can be achieved in your project? 
We'd love to hear your thoughts, ideas, insights, criticism etc. 

Thanks!

jongpie:
Fixes #183 for the unlocked package by introducing a new plugin called 'Filtered Log Entry Events' - I'm investigating options for making this plugin available for the managed package as well, but that will be handled in a separate pull request.


jongpie:
I'm closing this PR, and I'm not planning to move forward with this plugin - the Spring '22 release [provides beta-level support for filtering platform events](https://help.salesforce.com/s/articleView?id=release-notes.rn_platform_events_filtering_beta.htm&type=5&release=236) , so I don't think a custom plugin is needed anymore.

PR #187, opened 2021-08-24
Refactored trigger handler to manage more shared logic

jongpie:
Previously, each handler class was using `switch` statements around `Trigger.operationType` - the handler framework now manages it and calls protected methods as needed.


jamessimone (approved):
LGTM!

PR #188, opened 2021-08-26
Security enhancements

### LINKED ISSUE #198: Null Record causes error
After getting everything setup (thanks again!), I quickly ran into my first challenge. I have a Get Records, and immediately after it, I put a Add Log SObject action to log the results. But, the flow died with the action complaining it didn't get a required parameter: records.

Except I _did_ specify the parameter, it's just the results of the Get Records was nothing found, so the variable itself was empty, which is precisely what I wanted to log.

The absence of a parameter vs. the parameter being present and empty are two different things. Maybe Apex can't tell the difference in this circumstance(?), but if it can, it would be preferable if the action just logged the fact that the record was empty. I know I can put log actions on both sides of a test following the Get Records, but it would be better (cleaner) if that wasn't necessary.

Thanks!

### LINKED ISSUE #199: Log but no entries
Sorry, apparently I'm this week's problem child.

I have two "Add Log Entry for SObject Record Collection" actions in a flow, and only those actions. Both have Save Log set to True, both with Logging Level set to DEBUG. Both are in the path that will get called. (Confirmed with debug before I put the logging in.)

The Flow completes successfully. However, the Logger console shows me only a Log__c entry, no Log Entries. After the first time it happened, I deleted the log (so no logs at all) and ran the flow again, same result. Since I'm only calling actions that should create a log entry, I don't know how that could be happening; if the log actions weren't getting hit at all, there wouldn't be a log record, but since there's a log record, that implies that they are getting hit, but there's no log entry records, which the actions should both be creating.

I have the Logger: Admin permission set assigned to my user. (I'm a sysadmin.) I also created a Logger Setting entry for my user (and only that one, no org default), with Logging Level set to DEBUG and Is Enabled checked, everything else defaulted (Read, 14, EVENT_BUS, etc.). With my level set to DEBUG and the flow actions set to DEBUG, my understanding is that they should be created.

What am I missing and/or doing wrong?

jongpie:
This PR covers a few items that don't have issues logged, but handle a few security-related enhancements I've been considering for a few months (years??)

- :3rd_place_medal: **3rd place**: `LoggerAdmin` permission set now has access to all custom metadata types included in the repo. Is this a bugfix, or is this an enhancement? Who can say?! But all CMDT objects are now included.
- :2nd_place_medal: **2nd place**: new setting `LoggerSettings__c.StripInaccessibleRecordFields__c` - Disabled by default, when enabled, any time an `SObject` record is logged (or a `List<SObject>` is logged), only fields that the current user can access will be included in the record&apos;s JSON. This is useful in orgs where end-users have access to view `Log__c` and `LogEntry__c` records.
- :1st_place_medal: **1st place**: data masking for log entry messages. Enabled by default, regex-based CMDT rules can be configured to automatically mask sensitive data in the `Message__c` and `RecordJson__c` fields. Pre-built rules are included for credit card numbers and social security numbers - additional rules can be configured by creating additional `LogEntryDataMaskRule__mdt` records.

  - New `LoggerSettings__c` fields: `ApplyDataMaskRules__c` and `StripInaccessibleRecordFields__c`

    image

  - New custom metadata type `LogEntryDataMaskRule__mdt` list view, showing the included rules
  
    image

  - Example log entries with masked data - the 1st two entries shows masked credit card numbers, and the 3rd entry shows a masked social security number
    
    image

Other included changes:

- Fixed #198 Made record & records parameters optional in FlowRecordLogEntry & FlowCollectionLogEntry (respectively) to handle situations where null is passed (e.g., Get Records returns null for no matches, instead of an empty list)
- Fixed #199 - Changed FlowDescription__c.field to long text area + added automatic field-truncation for LogEntry__c inserts + added a ridiculously long description to one of the Flows in nebula-logger-recipes
- Switched to using custom getters for all constants in `LogEntryEventBuilder` that involve queries - this prevents the queries from executing if logging is disabled for the current user
- Added a private method `setTransactionDetails()` in `LogEntryEventBuilder` to defer setting some transactional details until absolutely necessary - this helps minimize heap size/avoid using some of the heap size limit until needed
- Updated Apex docs 
- Made fancier buttons that include the release number
- Enabled some user management settings in scratch definition files
- Cleaned up a few areas in README
- Added 'Code Quality Tests' to deploy.yml + ran prettier
- Setup ESLint for lwc + fixed some reported issues
- @jamessimone setup PMD rules & `apex-scanner` plugin + fixed some reported issues (via PR #201)
- Reenabled Codecov.io integration in pipeline
- Removed some unused scripts in package.json


jamessimone (approved):
@jongpie looks good to me! I went through and everything looks 👌 between the original and then incorporating the updates from #192 !

jongpie:
Thanks @jamessimone!! I think I'm going to first merge my other PR, #200, into this branch so I can release it all as 1 new package version, so I'm going to end up pushing a few more change to this PR & then I'll get it merged & released.

jamessimone:
Nice! Those will be nice to release together!

PR #200, opened 2021-09-10
Flow bugfixes + pipeline enhancements

jongpie:
Fixes 2 Flow-related issues reported by @vr8hub

- #198 - Made the `record` & `records` parameters optional in `FlowRecordLogEntry` & `FlowCollectionLogEntry` (respectively) to handle situations where null is passed (e.g., Get Records returns null for no matches, instead of an empty list)
- #199 - Changed field `LogEntry__c.FlowDescription__c` to a long text area field (instead of textarea) and added automatic field-truncation for LogEntry__c inserts
  - I originally created this field as a text area (255 characters) because the corresponding describe call says that a Flow's description can only be 255 characters, but it's all a lie! This Apex script will output `255` as the length, even though Flow descriptions can be much longer
    - `System.debug('FlowDefinitionView.Description max length is: ' + Schema.FlowDefinitionView.Description.getDescribe().getLength());`
  - Using a goofy-but-effective approach of simply adding a ton of text to a Flow's description, it looks like the true max length is 65,535. I've made the field 131,072 characters (the max length possible for a long text area field), just in case

I've also included a few enhancements for the repo's pipeline

- The pipeline now includes a "Code Quality Tests" node (in addition to "LWC Tests" and "Scratch Org Tests")
- Within the "Code Quality Tests" node, it runs:
  - ESLint for lightning web components
  - Prettier:verify for all code in the repo


vr8hub:
I'll just play devil's advocate for a moment: do we really need (potentially) 65K of flow description in the log entries? Knowing the permanent ID, the active version, the API version, etc. is useful. The description? Having 255 characters of it available is OK, but to me the rest seems excessive.

Just a random person's thoughts. :)

jongpie:
> I'll just play devil's advocate for a moment: do we really need (potentially) 65K of flow description in the log entries? Knowing the permanent ID, the active version, the API version, etc. is useful. The description? Having 255 characters of it available is OK, but to me the rest seems excessive.
> 
> Just a random person's thoughts. :)

Hey @vr8hub - I see what you're saying, and honestly, I'm surprised Salesforce even supports storing that long of a description for Flow. But I try to design the data model for Logger to match/accommodate whatever data is being logged from the platform, and minimize truncation as much as possible. Switching the field to a long text area (instead of 255 text area) doesn't really have any major negative impacts (as far as overall performance, storage usage, etc), so I'd rather increase the field to handle the possible data, in case other people do have a strong need to log the full Flow description.

I'm hoping to wrap up this PR today - it'll be included in the next release, v4.6.7. Thanks again for all of the feedback!

PR #201, opened 2021-09-13
Added static code analysis / PMD

### LINKED ISSUE #194: Setup apex scanner in the pipeline
To provide static code analysis, the [sfdx scanner plugin](https://forcedotcom.github.io/sfdx-scanner/) should be added to the pipeline so that any issues found will fail the build. Not all of the included rules will apply to Logger, so they'll need to be customized.

jamessimone:
* Fixes #194 by adding PMD through `sfdx-scanner`

### DIFF
--- a/nebula-logger/tests/log-management/classes/LogMassDeleteExtension_Tests.cls
+++ b/nebula-logger/tests/log-management/classes/LogMassDeleteExtension_Tests.cls
@@ -113,7 +113,7 @@ private class LogMassDeleteExtension_Tests {
 
             String deleteAccessError = 'You do not have access to delete logs records';
 
-            LogMassDeleteExtension extension = new LogMassDeleteExtension(controller);
+            new LogMassDeleteExtension(controller);
 
             System.assertEquals(true, ApexPages.hasMessages(ApexPages.SEVERITY.ERROR));
             System.assertEquals(deleteAccessError, ApexPages.getMessages().get(0).getSummary());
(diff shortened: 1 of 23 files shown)


nebula-logger/tests/logger-engine/classes/LogEntryEventBuilder_Tests.cls
```diff
@@ -887,7 +885,7 @@ private class LogEntryEventBuilder_Tests {
         // Get expected data
         Organization organization = LoggerTestUtils.getOrganization();
         String organizationEnvironmentType = LoggerTestUtils.getOrganizationEnvironmentType();
-        SObject networkSite = LoggerTestUtils.getNetwork();
+        LoggerTestUtils.getNetwork();
```
jongpie, line 888: I think this line can just be removed, it doesn't seem like it's being used

nebula-logger/tests/log-management/classes/LogMassDeleteExtension_Tests.cls
```diff
@@ -113,7 +113,7 @@ private class LogMassDeleteExtension_Tests {
 
             String deleteAccessError = 'You do not have access to delete logs records';
 
-            LogMassDeleteExtension extension = new LogMassDeleteExtension(controller);
+            new LogMassDeleteExtension(controller);
```
jongpie, line 116: This seems like this line could be removed altogether think?

jamessimone: this one is necessary, otherwise the error doesn't get successfully loaded to the page

jongpie: Doh! That makes sense, ignore me!

jongpie (approved):
Looks great to me, it's ready to merge! Thanks again for working on this!

PR #218, opened 2021-10-06
New 'Scenario' field on Log__c

### LINKED ISSUE #217: Add new 'scenario' field to group logs of the same context and easily find the wanted log
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/207

<sup>Originally posted by **fentes** September 17, 2021</sup>
Hey,

we introduced NebulaLogger in our Org and are amazed by the easy use. So we use it in many different places. Unfortunately, this leads to the fact that we have a lot of logs and it is, therefore, difficult to find the right log only by the username and the time.

To keep the overview, we changed some lines and introduced a new field "Scenario". The _Logger.saveLog()_ was cloned into _Logger.saveLog(String myScenario)_

image

I think this might be quite helpful for others as well ;) </div>

jongpie:
Closes #217 by adding a new field, `Log__c.Scenario__c`, that can be set via Apex, Flow and lightning components. The scenario field provides a way to identify the context (or.... dare I say, the scenario) of the current transaction. This field is provided as a way for org's to self-identify the context of the current transaction. It fully relies on the org's metadata to handle setting this - nothing within Nebula Logger automatically sets this field. In the event that the scenario is set multiple times in a single transaction, then the last scenario specified will be the value used for `Log__c.Scenario__c`.

|        | Scenario Field                                  | Tagging System                                                      |
| ------ | ----------------------------------------------- | ------------------------------------------------------------------- |
| Object | `Log__c` object - stored in `Log__.Scenario__c` | `LogEntry__c` object - related via `LogEntryTag__c` junction object |
| Usage  | Only 1 scenario can be stored per transaction   | Supports adding multiple tags to each `LogEntry__c` record          |

## Specifying Scenario in Apex
When logging from within Apex, simply call the method `Logger.setScenario(String scenario)` at some point within your code to specify the current transaction's scenario.

```java
Logger.info('some log entry created BEFORE specifying a scenario'); // This log entry will still have the scenario set, even though it's logged before `setScenario(String)` is called
Logger.setScenario('an example transaction scenario name');
Logger.info('some log entry created AFTER specifying a scenario'); 
Logger.saveLog();
```

## Specifying Scenario in Lightning Components
When logging from within lightning web components and aura components, you can specify the scenario by calling the `logger` function `setScenario(scenario)` - this function expects a String to be passed as the parameter. This example shows a basic example of how you can specify the log's scenario and save any pending log entries.

```javascript
saveLogExample() {
    const logger = this.template.querySelector(LOGGER_NAME);
    logger.setScenario(this.scenario);
    console.log(logger);
    logger.saveLog();
}
```

## Specifying Scenario in Flow
With adding a log entry from Flow, you can now optionally set the transaction's scenario, using the provided property `Scenario` - this property is available in all 3 Flow classes (`FlowLogEntry`, `FlowRecordLogEntry`, and `FlowCollectionLogEntry`).

image

## Scenario Field in `Log__c` List Views
The new `Scenario__c` field is now included in the "All Logs" list view - you can also add the new field to any of your own list views.

image

## Scenario Field on `Log__c` Detail Page
The new `Scenario__c` field is also included on the `Log__c` record's detail page, under the section "Log Management"

image

## Manually Updating the Scenario Field via Log's "Manage" Quick Action
In some situations, you may want to manually change the value of the field `Log__c.Scenario__c` - this can now be done using the "Manage" quick action. Simply click the "Manage" button on a `Log__c` record (or select multiple logs from a list view and click "Manage"), and you can then update the field's value.

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,8 +3,6 @@ name: Deployment
 
 on:
     push:
-        branches:
-            - main
         paths-ignore:
             - 'content/**'
             - 'docs/**'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,8 @@ name: Deployment
 
 on:
     push:
+        branches:
+            - main
         paths-ignore:
             - 'content/**'
             - 'docs/**'
--- a/docs/logger-engine/ComponentLogger.md
+++ b/docs/logger-engine/ComponentLogger.md
@@ -104,6 +104,10 @@ The value to use as the log entry&apos;s message
 
 (Optional) The record ID to relate to the log entry
 
+###### `scenario` → `String`
+
+Optionally specify the name to use for the current transaction&apos;s scenario
+
 ###### `stack` → `String`
 
 The JavaScript stack trace from when the log entry was created
--- a/docs/logger-engine/FlowCollectionLogEntry.md
+++ b/docs/logger-engine/FlowCollectionLogEntry.md
@@ -46,6 +46,10 @@ The records to relate to this log entry - the records&apos; JSON is automaticall
 
 Optionally choose to save any pending log entries
 
+#### `scenario` → `String`
+
+Optionally specify the name to use for the current transaction&apos;s scenario
+
 #### `tagsString` → `String`
 
 Optionally provide a comma-separated String of tags to dynamically assign to the log entry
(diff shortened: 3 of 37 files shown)


fentes:
@jongpie Thank you very much for implementing this feature into the main framework. This will really help to keep the overview of all Logs and let us easily find the wanted log. 
I skimmed through the code changed and it looks good. Happy to test this in our org.

fentes:
Hey installed the beta package on a sandbox and it works fine 😁 Let me now, if you have a release package.

jongpie:
@fentes that's great news! I'm going to include 1-2 other unrelated changes in this release, and I hope to have it completed by early next week. I'll let you know as soon as it's ready!

jongpie:
@jamessimone thanks for reviewing this!

Issue #212, opened 2021-09-22
Log Entry Event Stream enhancements


jongpie:
### New Feature Summary

In release [v4.6.8](https://github.com/jongpie/NebulaLogger/releases/tag/v4.6.8), I introduced a new lwc `logEntryEventStream` to monitor & filter `LogEntryEvent__e` platform events in real time (PR #191). But, there are still several enhancements that I'd like to make that weren't included in the original release:

- Add support for using regex when filtering on the 'message' field
- Finish implementing the commented-out "expand/collapse" button. I implemented the basics for this functionality, but it had some issues on different resolutions
- Make the left-side panel (Stream Settings) collapsible, using SLDS's [split view](https://www.lightningdesignsystem.com/components/split-view/)
- Add a way to toggle between the current 'console' view and a datatable view

PR #251, opened 2021-11-10
Regex filtering in logEntryEventStream component

jamessimone:
* Fixes (some) of #212 by adding regular expression filtering to the `logEntryEventStream` component. Did some cleanup, as well

### DIFF
--- a/nebula-logger/main/log-management/lwc/logEntryEventStream/logEntryEventStream.js
+++ b/nebula-logger/main/log-management/lwc/logEntryEventStream/logEntryEventStream.js
@@ -2,6 +2,7 @@ import { LightningElement } from 'lwc';
 import { isEmpEnabled, subscribe, unsubscribe } from 'lightning/empApi';
 
 export default class LogEntryEventStream extends LightningElement {
+    unfilteredEvents = [];
     logEntryEvents = [];
     isExpanded = false;
     isStreamEnabled = true;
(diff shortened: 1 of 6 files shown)


nebula-logger/main/log-management/lwc/logEntryEventStream/logEntryEventStream.js
```diff
@@ -104,44 +97,38 @@ export default class LogEntryEventStream extends LightningElement {
 
     onToggleStream() {
         this.isStreamEnabled = !this.isStreamEnabled;
-        if (this.isStreamEnabled) {
-            this.createSubscription();
-        } else {
-            this.cancelSubscription();
-        }
+        // eslint-disable-next-line
+        this.isStreamEnabled ? this.createSubscription() : this.cancelSubscription();
```
jamessimone, line 101: let me know what you think - when I wrote this, I thought it was nice, but then eslint hated it so ...

jongpie: Yeah, I like this approach - I also think it looks nice, so let's stick with this.

nebula-logger/main/log-management/lwc/logEntryEventStream/logEntryEventStream.js
```diff
@@ -68,24 +75,9 @@ export default class LogEntryEventStream extends LightningElement {
...
-        this.originLocationFilter = event.target.value;
-    }
-
-    handleLoggedByFilterChange(event) {
-        this.loggedByFilter = event.target.value;
-    }
-
-    handleMessageFilterChange(event) {
-        this.messageFilter = event.target.value;
+    handleFilterChange(event) {
```
jongpie, line 78: Awww yeah, I love this change - having only 1 common `change` function is much better 🚀

jongpie (approved):
This looks awesome - thanks again for making these changes!

PR #254, opened 2021-11-30
New 'Logger Settings' lwc/tab

### LINKED ISSUE #202: Add a component for managing LoggerSettings__c records
Currently, Logger relies on the standard Salesforce UI for managing `LoggerSettings__c`, but I think it'd be nice to include an lwc (possibly displayed as a tab in Logger Console) that provides the ability to manage the records. This would provide 2 benefits

- **Better UI/UX**: custom settings don't support picklist fields, but some of the fields on `LoggerSettings__c` only work with predefined values (examples: `LoggingLevel__c`, `DefaultSaveMethod__c`, etc.). Using a custom component could support displaying these fields as picklist values.
- **Providing support to edit `LoggerSettings__c` for non-system admins**: in order to update custom settings records in Salesforce, you have to have "Customize Application" access, and some logger admins may not have this access in their prod org. By using a custom component, logger admins will be able to edit `LoggerSettings__c` records without granting the "Customize Application" permission.

jongpie:
Closes #202 by adding a new lwc, `loggerSettings`, as a tab in the Logger Console app. It provides several features for managing `LoggerSettings__c` records:

- Provides a datatable of all `LoggerSettings__c` records - long-term, more optimizations may be needed for orgs that have a lot of records (currently, there is infinite scrolling or pagination)
  
  image

- Provides view, create, edit and delete functionality. Anyone with access to the tab/lwc can see the datatable and the 'View' row action - but in order to access/use the 'New' button (create), and the 'Edit' & 'Delete' row actions, users must either have update access on the `LoggerSettings__c` object(`LoggerSettings__c.SObjectType.getDescribe().isUpdateable()`), or have the custom permission 'Can Modify Logger Settings' (`FeatureManagement.checkPermission('CanModifyLoggerSettings')`). This is useful for orgs where people may have the `LoggerAdmin` permission set, but aren't full System Admins.
  - The `LoggerAdmin` permission set has access to view, create, edit, and delete
  - The `LoggerLogViewer` permission set only has access to view
  - The `LoggerLogCreator` and `LoggerEndUser` permission sets do not have access to the tab/lwc
  
  image

- Provides page sections to mimic the look & feel of standard page layouts. Custom settings objects do not support page layouts, so there's normally not a way to control the arrangement of fields show in the UI. Within the component, all of the fields have been arranged into 5 sections:
  - General Settings
  - Log Management Settings
  - Developer Settings
  - Security Settings
  - System Information
  
  image

- Displays usernames (instead of a user's _name_) as the primary field for user-specific records. This is useful in orgs where multiple users may have the same first/last name - with the standard UI for custom settings, it can be difficult to select the correct user when multiple people have the same name. Since username is unique, the component provides a better method of displaying/selecting the appropriate user.

  image

- Provides picklist-style inputs for fields that really should be picklists. Custom settings objects do not support several field types, including picklists - instead, text fields are used on the object, but this can lead to bad config data being entered (e.g., an admin/developer mistypes the name of a `LoggingLevel` or `Logger.SaveMethod` enum value. The lwc solves this by presenting these fields as picklists, using some Apex in the `LoggerSettingsController` class to generate the list of picklist options. This prevants any isssues with tpyos, and it simplifies the configuration process for admins/developers

  image

- Field labels updated: previously, I had shortened the labels on several fields to try to make things better fit into the standard list views for `LoggerSettings__c`. I've re-expanded several of the labels to be more verbose/descriptive, since the component provides a better use of space
- Bonus unexpected feature: using this component, you can now configure `LoggerSettings__c` for the the Automated Process user. When using the standard UI for custom settings, you cannot search for/select the Automated Process user, but it apparently works just fine when the settings record is created via Apex
  
  image

---

## Other changes
I've also included some other changes in this PR:
- Pipeline enhancements - I've added caching for `sfdx-cli` in the pipeline, and the pipeline now deploys to 2 different scratch orgs (in parallel) to confirm the metadata works in both a base org, and an experience cloud org
- Documentation enhancements - I've added JSDoc comments for the `logger` lwc classes - the GitHub Pages site should now include these docs going forward (alongside the existing Apex docs). I also cleaned up a few ApexDoc comments in various classes


nebula-logger/main/log-management/lwc/loggerSettings/loggerSettings.js
```diff
@@ -0,0 +1,370 @@
...
+    // LoggerSettings__c data
+    records;
+    currentRecord;
+    selectedSetupOwner;
+
+    connectedCallback() {
+        document.title = this.title;
+        this.showLoadingSpinner = true;
+
+        getOrganization().then(result => {
```
jamessimone, line 60: slight optimization possible here - you could wrap each of your `thennable` promises and then call `Promise.all` on an array of those promises so that they can run in parallel.

jongpie: Ooooh, that's a great idea - I'll make that change ASAP. Thanks for taking a look!

jongpie: I've optimized this in [this commit](https://github.com/jongpie/NebulaLogger/pull/254/commits/8d48fa11d8d1996849f67727593669d363bc2152) by consolidating all of the different Apex methods into a single method that returns all of the various picklist options in a single object

jamessimone: Makes sense - and as you said, what a speed up!

jongpie:
@jamessimone, I.... I _think_ this PR is finally ready!...? If you have some more time to spare, would you mind taking another look before I merge?

jamessimone:
> @jamessimone, I.... I _think_ this PR is finally ready!...? If you have some more time to spare, would you mind taking another look before I merge?

Absolutely. I’ll have a look through this morning!

nebula-logger/main/log-management/classes/LoggerSettingsController.cls
```diff
@@ -0,0 +1,372 @@
...
+        public LoggerSettings__c record;
+
+        /**
+         * @description Implements sorting logic for the `Comparable` interface
+         * @param  compareTo The object to compare to when sorting the current item
+         * @return           The sorting result
+         */
+        public Integer compareTo(Object compareTo) {
+            SettingsRecordResult that = (SettingsRecordResult) compareTo;
+            String thisCompositeName = this.setupOwnerType + '.' + this.setupOwnerName;
```
jamessimone, line 345: you can encapsulate the composite name in a private method and call it in `compareTo`:

```java
private String getCompositeName() {
  return this.setupOwnerType + '.' + this.setupOwnerName;
}

// and then in compareTo:
if (this.getCompositeName() == that.getCompositeName() {
  // etc ...
}
```

Maybe it's not worth it in this exact scenario, but I thought I'd mention it in case you like that better.

nebula-logger/main/log-management/lwc/loggerSettings/__tests__/loggerSettings.test.js
```diff
@@ -0,0 +1,551 @@
...
+        }
+        jest.clearAllMocks();
+    });
+
+    it('loads component without modify access', async () => {
+        const loggerSettingsElement = await initializeElement(false);
+
+        // Check the component
+        expect(document.title).toEqual('Logger Settings');
+        // expect(getObjectInfo.getLastConfig()).toEqual(mockObjectInfo);
```
jamessimone, line 145: remove this

nebula-logger/main/log-management/lwc/loggerSettings/__tests__/loggerSettings.test.js
```diff
@@ -0,0 +1,551 @@
...
+    });
+
+    it("shows record modal when 'new' button is clicked", async () => {
+        const loggerSettingsElement = await initializeElement(true);
+        createRecord.mockResolvedValue(mockNewRecord);
+
+        // Clickety click
+        const newRecordBtn = loggerSettingsElement.shadowRoot.querySelector('lightning-button[data-id="new-btn"]');
+        newRecordBtn.click();
+        // Yes, 2 calls for Promise.resolve() are needed... because of... reasons?
```
jamessimone, line 182: I really love this comment

nebula-logger/tests/log-management/classes/LoggerSettingsController_Tests.cls
```diff
@@ -0,0 +1,378 @@
...
+        List<String> expectedSetupOwnerType = new List<String>{ 'Organization', 'Profile', 'User' };
+        Integer expectedSetupOwnerTypeSize = 4; // Options are org, profile, and user, and '--NONE--' is automatically included
+        List<LoggerSettingsController.PicklistOption> picklistOptions = LoggerSettingsController.getPicklistOptions().setupOwnerTypeOptions;
+        System.assertEquals(expectedSetupOwnerTypeSize, picklistOptions.size());
+        for (LoggerSettingsController.PicklistOption picklistOption : picklistOptions) {
+            if (String.isBlank(picklistOption.value) == true) {
+                System.assertEquals('--None--', picklistOption.label);
+            } else {
+                System.assertEquals(picklistOption.value, picklistOption.label);
+                System.assertEquals(true, new Set<String>(expectedSetupOwnerType).contains(picklistOption.value));
```
jamessimone, line 130: any reason you're converting to a set here? if this was in production-level code and it was a big list, I wouldn't use `List.contains()` but for a test with only a few strings in it, I think it's fine. (same with the below test(s))

nebula-logger/tests/log-management/classes/LoggerSettingsController_Tests.cls
```diff
@@ -0,0 +1,378 @@
...
+        List<LoggerSettings__c> testSettingsRecords = new List<LoggerSettings__c>{ orgDefaultSettingsRecord, profileSettingsRecord, userSettingsRecord };
+        insert testSettingsRecords;
+        Map<Id, LoggerSettings__c> testSettingsRecordsById = queryLoggerSettingsRecords(testSettingsRecords);
+
+        List<LoggerSettingsController.SettingsRecordResult> recordResults = LoggerSettingsController.getRecords();
+        System.assertNotEquals(true, recordResults.isEmpty());
+        System.assertEquals(testSettingsRecordsById.size(), recordResults.size());
+        for (LoggerSettingsController.SettingsRecordResult result : recordResults) {
+            LoggerSettings__c matchingRecord = testSettingsRecordsById.get(result.record.Id);
+            // String expectedSetupOwnerType = matchingRecord.SetupOwner.Type;
```
jamessimone, line 179: remove these comments

nebula-logger/main/log-management/classes/LoggerSettingsController.cls
jongpie: Oooh, I do like that approach, I'll add that private method - it seems like a cleaner way to organize the code.

nebula-logger/main/log-management/lwc/loggerSettings/__tests__/loggerSettings.test.js
jongpie: It turned out to be a necessary comment 😆 - since I was working on this PR for a while, I kept forgetting that I _do_ need both, so I kept removing one of the lines on different days

nebula-logger/tests/log-management/classes/LoggerSettingsController_Tests.cls
jongpie: I think these lists are all small enough that I'm not too worried about using `.contains()`.... but I'm not sure why I defined the variables as `List<>` and then just convert to `Set<>`, that part seems silly. I'll change the variables to be initially defined as `Set<>`

PR #274, opened 2022-01-20
Spring '22 release - v4.7.0

### LINKED ISSUE #15: Add wiki for documentation

### LINKED ISSUE #195: Logger in Flow questions
I've read the Readme multiple times and browsed the fields in the Log and LogEntry objects, but I'm still missing some information re how to use this in Flow. (No Apex development here, so this will be used purely in Flow.)

1. What is the optional "Save Log" field on the "Add Log Entry" action for, i.e. when would we use it or not, and if we do use it, what value are we putting in the field? (An additional screen print in the flow quick start that showed the parameters for that action might be helpful.)
2. A similar question for the "Save Log" action—the readme says it "saves any pending logs," but what makes a pending log? Do the other three "Add" actions only create pending logs and we need to use a "Save Log" each time we want to save? That doesn't sound right, but there's no other mention of "pending logs" re Flow, so I'm not groking what we use that action for.
3. The Add Log event takes a record ID, the "… for SObject" takes a record variable, and the "… for Sobject Record Collection" a collection variable, but what is the end-result difference in those? I didn't see a record ID field on the Log__c object, so what is Add Log Event doing with it? Are the other two actions actually storing the record(s) in LogEntry? I'm trying to get a handle on why we would use one vs. the other.

Thanks so much, and I apologize in advance if I missed where these were addressed.

### LINKED ISSUE #258: Toggle to not save logs to the object and write to Apex Debug logs
### New Feature Summary

Add a `LoggerSettings` to turn on / off the saving of logs to logging custom objects. When turned off, it should write the log to Apex debug log instead so that we can use the standard debug log view to view the logged info.

### LINKED ISSUE #263: Use the new package ancestor keywords when generating managed package versions
The Spring '22 release adds some new keywords that can be used for specifying a managed package's ancestor dynamically - logger should start using [the `HIGHEST` keyword](https://help.salesforce.com/s/articleView?id=release-notes.rn_packaging_ancestor_enhancements.htm&type=5&release=236) in `sfdx-project.json` so that it doesn't have to be updated for every release.

### LINKED ISSUE #271: Unit tests are failing, when running user's language is not English

Unlocked Package

4.6.14

### New Bug Summary

We are using NebulaLogger in several projects successfully. Unfortunately several unit tests are failing - all we the same root cause:

`System.QueryException: List has no rows for assignment to SObject`

The reason can be found here:

https://github.com/jongpie/NebulaLogger/blob/f190ab12b19baa14d78921d929b3c8a57f20a93b/nebula-logger/core/tests/common/classes/LoggerTestUtils.cls#L10

This query - using a profile name as where clause - will fail, if the running user's language is not English. In our Orgs the user uses usually German as language, so the profile name will be "Standardbenutzer" instead of "Standard User". So this query fails. 

Workaround:
* Switch running user's language to English

### LINKED ISSUE #272: Nebula Logger: Illegal assignment from List<Schema.PicklistEntry> to List<PicklistEntry>
Hi,

when trying to install Nebula Logger I get the following error:
```
1.  Illegal assignment from List<Schema.PicklistEntry> to List<PicklistEntry>
LogHandler: Illegal assignment from List<Schema.PicklistEntry> to List<PicklistEntry>

2.  Method does not exist or incorrect signature: void getValue() from the type PicklistEntry
LogHandler: Method does not exist or incorrect signature: void getValue() from the type PicklistEntry

3.  Method does not exist or incorrect signature: void getValue() from the type PicklistEntry
LogHandler: Method does not exist or incorrect signature: void getValue() from the type PicklistEntry

4.  Dependent class is invalid and needs recompilation:
 Class LogHandler : Illegal assignment from List<Schema.PicklistEntry> to List<PicklistEntry>
LogHandler_Tests: Dependent class is invalid and needs recompilation:
 Class LogHandler : Illegal assignment from List<Schema.PicklistEntry> to List<PicklistEntry>

5.  Invalid type: LogHandler
Log: Invalid type: LogHandler

6. (LoggerLogManagement) No classes found for LogHandler_Tests.
LoggerLogManagement: No classes found for LogHandler_Tests.
```

I have actually noticed that Salesforce has recently changed the return type of getPicklistValues() and started to throw type mismatch on Schema vs global PicklistEntry.

I can make PR if needed :) 

Thank you!

### LINKED ISSUE #276: LoggerEmailUtils_Tests failed

Unlocked Package

https://github.com/jongpie/NebulaLogger/releases/tag/v4.6.16

### New Bug Summary

After a refresh of the full and partial sandbox two tests are failing.

```
System.EmailException: SendEmail failed. First exception on row 0; first error: NO_MASS_MAIL_PERMISSION, Single email is not enabled for your organization or profile.: []
--
Stack trace: | Class.LoggerEmailUtils.sendEmail: line 111, column 1 Class.LoggerEmailUtils.sendErrorEmail: line 81, column 1 Class.LoggerEmailUtils.sendErrorEmail: line 63, column 1 Class.LoggerEmailUtils_Tests.it_should_send_email_notification_for_upsertResult_errors_when_enabled: line 63, column 1
```

```
System.EmailException: SendEmail failed. First exception on row 0; first error: NO_MASS_MAIL_PERMISSION, Single email is not enabled for your organization or profile.: []
--
Stack trace: | Class.LoggerEmailUtils.sendEmail: line 111, column 1 Class.LoggerEmailUtils.sendErrorEmail: line 81, column 1 Class.LoggerEmailUtils.sendErrorEmail: line 47, column 1 Class.LoggerEmailUtils_Tests.it_should_send_email_notification_for_saveResult_errors_when_enabled: line 16, column 1
```

### LINKED ISSUE #279: Add method Logget.getScenario() and new DefaultLogScenario__c field on LoggerSettings__c
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/278

<sup>Originally posted by **dschibster** February  6, 2022</sup>
Hi there @jongpie!

I am in the middle of cleanup project, and want to get to use Nebula Logger to finally get some transparency about when automation go awry. One thing I want to use extensively is the scenario feature. However, I think that when you enter Apex from a Component, it's pretty easy to set one, but not so much when you run a basic update from UI. The thing is, I was thinking if setting a Default Scenario in these cases would be wise, which I wouldn't be able to do right now because I wouldn't be able to inspect the Scenario at runtime without fiddling in the code.

What do you think about the option of adding a Default Scenario, and what would you say about adding a getter to check if we need to set one in our Triggers?</div>

jongpie:
## New Features for `LoggerSettings__c`
- Closed #258 by adding 2 new `LoggerSettings__c` fields to provide control over platform events & custom objects
  1.   `IsSavingEnabled__c` - when disabled, any calls to Logger.saveLog() are ignored. This allows you to still use Logger for outputting Apex debug statements, but it completely disables saving/publishing of the platform event `LogEntryEvent__e`.
  2. `IsPlatformEventStorageEnabled__c` - this fields controls if `LogEntryEvent__e` platform events are transformed & stored in the custom objects `Log__c` and `LogEntry__c` (when `IsSavingEnabled__c` == true).
- Closed #279 by adding some new features for "log scenarios"
  - Configure a default scenario by setting the new field `LoggerSettings__c.DefaultLogScenario__c` 
  - Retrieve the current transaction's scenario in Apex using new method `Logger.getScenario()`
- Added the ability to assign a default `OwnerId` for new `Log__c` records via a new settings field, `LoggerSettings__c.DefaultLogOwner__c`. This is useful for orgs that assign/manage logs for triage purposes (often done in production environments for managing logs with errors & warnings) and want more control over who is assigned as the log owner. The setting can be configured with 1 of 5 possible values
  1. Blank (null) - this will provide the same behavior that has existed in previous releases - the `Log__c` record will simply be assigned to the user that generated the log
  2. User ID
  3. Username
  4. Queue ID
  5. Queue DeveloperName 
- Renamed `LoggerSettings__c` field `StripInaccessibleRecordFields__c` to `IsRecordFieldStrippingEnabled__c` for a consistent field naming convention
- Updated lwc `loggerSettings` and `logEntryEventStream` to support using a namespace when running in the managed package by introducing a new Apex class, `LoggerSObjectMetadata`. This class provides LWCs with info about `SObject`, `SObjectField`, and `PicklistEntry` metadata (including details about a namespace, when applicable).
- Bumped all metadata to API v54.0 (Spring '22 release)
- Added picklist values in `Log__c.ApiVersion__c` for all 3 API versions in calendar year 2022
- Closed #15 and closed #195 by expanding wiki content. Changes for the wiki are managed in a separate repo, so this PR doesn't contain the changes, but the wiki updates have been done in parallel

image

## New Features for `Log__c` Quick Action "Open Viewer" (formerly "View JSON")
- Renamed the lwc `logJSON` _back_ to `logViewer`. This change only impacts the unlocked package.
- Renamed the quick action "View JSON" (again) to "Open Viewer"
- Enhanced `logViewer` lwc to display tabs for both JSON (existing) and log file (new). The new 'log file' view displays the log's related log entries in a streamlined view that includes each entry's timestamp, logging level, message, and stack trace
- Added a new download button to `logViewer` to export the selected content to a file

image

## Bugfixes
- Fixed #271 - Updated `LoggerTestUtils` to use the current user's profile in most cases, instead of trying to query for a particular profile. For testing behavior related to profiles without access to Logger, integration tests have been added to the `extra-tests` folder that still leverage the 'Standard User' profile in a scratch org, but these tests are not included in the package.
- Fixed #272 by explicitly declaring the list as `List<Schema.PicklistEntry>` instead of just `List<PicklistEntry>`
- Fixed #276 by updating `LoggerEmailUtils` to check if email deliverability is enabled in the current org prior to attempting to send any emails
- Wrapped all dynamic query strings in `String.escapeSingleQuotes()` as a security best practice & to avoid PMD reporting errors for `ApexSOQLInjection`

Pipeline / DevOps
- Updated pipeline to use JWT bearer flow + sfdx bot user for dev hub auth
- Updated the Experience Cloud scratch definition to use Spanish as the language/Spain as the country as another way to monitor/test for any language-related issues in the pipeline
- Cleaned up some PMD rule suppressions
- Added basic Experience Cloud site to the pipeline to provide a better method of testing Experience Cloud scenarios (the site is not included in the package itself - it's only used in the pipeline). Previously, the pipeline created an Experience Cloud site via the async function `sfdx force:community:create` + a 2 minute delay, but the async operation is inconsistent in how long it takes to complete. By deploying the site metadata in the pipeline, it becomes a synchronous operation, which should prevent some inconsistent pipeline failures.
- Updated/added some `devDependencies` in `package.json` to handle some dependabot alerts
- Closed #263 by adding `"ancestorVersion": "HIGHEST"` in the managed package's `sfdx-project.json` file


nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls
```diff
@@ -56,6 +56,13 @@ public without sharing class LogEntryEventHandler extends LoggerSObjectHandler {
         Log__c recentLogWithApiReleaseDetails = getRecentLogWithApiReleaseDetails();
 
         for (LogEntryEvent__e logEntryEvent : this.logEntryEvents) {
+            // Load the logging user's settings
+            User loggingUser = new User(Id = logEntryEvent.LoggedById__c, ProfileId = logEntryEvent.ProfileId__c);
```
jamessimone, line 60: seems like this could be consolidated with the below addition

jongpie:
@jamessimone thanks for reviewing! I'll incorporate your comments ASAP

PR #275, opened 2022-01-25
Automatically enable platform event saving when general saving is enabled

jamessimone:
* Automatically check platform event saving input if general Enable Saving checkbox is clicked


jamessimone:
Woohoo! Thank you!

Issue #128, opened 2021-03-31
Add support for archiving Log__c and LogEntry__c records into a BigObject


jongpie:
Related to #117 - for orgs that store records in `Log__c` and `LogEntry__c`, there should be a way to archive records as an alternative to deleting them.

Proposed Changes
* **New field `Log__c.LogPurgeAction__c`** - this should be a picklist with 2 values, 'Delete' & 'Archive'. `LogBatchPurger` should be updated to look at this field and based on the field's value, it will then delete the `Log__c` record (current functionality) or archive it to the new BigObject (tentatively called `LogEntryArchive__b` for now). 
  * This field should also be added to the 'Manage' quick action on `Log__c`
  * `LogRetentionDate__c` will continue to control _when_ the `Log__c` is purged - `LogPurgeAction__c` will then control _how_ the `Log__c` is purged (delete vs archive)
  * `LoggerSettings__c` should also have a new text field - `DefaultLogPurgeAction__c` - to control the default value when creating the `Log__c` record in `LogEntryEventHandler`
* **Create a new quick action called 'Archive' on `Log__c`** - when clicked, it should look similar to the modal shown for the standard 'Delete' button, and then archive the record when the user confirms. This might require creating an `aura` component - I don't think Flows would give us full control over the UI, and `lwc` still are not supported as quick actions.
* **Create a new 'Mass Archive' list button on `Log__c`** - it should function similar to the 'Mass Delete' button (but archive the selected log records, instead of deleting them)

Somewhat in reverse to how `LogEntryEvent__e` records are normalized into `Log__c` and `LogEntry__c` records, when archived, the `Log__c` and `LogEntry__c` records should be denormalized back into a single record and stored in `LogEntryArchive__b`

PR #288, opened 2022-03-22
Added support for plugins within LogBatchPurger + added archiving in BigObject plugin

jongpie:
As a follow-up to #287 (and part of a larger overhaul of the plugin framework for the upcoming release `v4.7.1`), I've reworked even more of the plugin framework to add support for plugins within the `LogBatchPurger` class

## Plugin Framework Changes
- Eliminated the old plugin approach by deleting the abstract class `LoggerSObjectHandlerPlugin`. It has been replaced with 2 new interfaces:
  1. `LoggerPlugin.Batchable` - used to create plugins for `LogBatchPurger` (batch deletion job)
  2. `LoggerPlugin.Triggerable` - used to create plugins for `LoggerSObjectHandler` (trigger framework used by Logger)
- Cleaned up the fields on `LoggerPlugin__mdt` + added new fields to control the sort order of plugins used by `LogBatchPurger`
- Added a new settings field `LoggerSettings__c.DefaultLogPurgeAction__c` - this then sets the default value of the field `Log__c.LogPurgeAction__c`. This field value isn't used by `LogBatchPurger` directly, but it provides a place for plugins to set/read additional actions that can be used within plugin logic (see BigObject changes below for an example)
  - I've also updated `LoggerSettingsController` to query for `LoggerParameter__mdt` records that start with 'CustomLogPurgeAction' so that plugins can define & package new actions. Any matching values will be displayed in the LWC `loggerSettings` as a picklist option for `LoggerSettings__c.DefaultLogPurgeAction__c`
- `LogBatchPurger` now uses a new inner class `BatchPurgerInput` to provide contextual info to plugins
  - I realized earlier today that the class `LogBatchPurger.BatchPurgerInput` and `LoggerSObjectHandler.SObjectHandlerInput` will both need to be changed to top-level classes so that Flow plugins can use them. From [this page](https://help.salesforce.com/s/articleView?id=sf.flow_considerations_apex_data_type.htm&type=5), Apex-defined types can't be used in Flow unless they are top-level classes. I've already scope-creeped in this branch a lot, so I'll make this change after merging this PR into the `plugin-framework-overhaul` branch.
- Scope creep: standardized some of the `@TestVisible` methods used throughout the codebase to mock custom metadata records

## BigObject Plugin Changes
The BigObject plugin has been updated to also work as a batch purger plugin (in addition to being an SObject handler plugin that @jamessimone implemented in #287)
- Split the class `LogEntryArchiveBuilder` into 2 classes:
  1. `LogEntryArchiveBuilder` - this now only handles converting a `LogEntryEvent__e` into a `LogEntryArchive__b` record (implemented via #287), as well as converting a `LogEntry__c` record into a `LogEntryArchive__b` record (to cover most of the functionality in #128)
  2. `LogEntryArchivePlugin` - this new class implements the interfaces `LoggerPlugin.Batchable` and `LoggerPlugin.Triggerable` so archiving can be used as the initial storage place (when a user's save method == `BIG_OBJECT`) as well as a secondary storage place (when a `Log__c` has `LogPurgeAction__c` == 'Archive')
- Added more fields on `LogEntryArchive__b` to (more or less) match the list of fields on `LogEntryEvent__e`
- The BigObject plugin now includes a custom value 'Archive' for the field `LoggerSettings__c.DefaultLogPurgeAction__c`. When `LogBatchPurger` executes, any matching `Log__c`/`LogEntry__c`/`LogEntryTag__c` data will be archived into `LogEntryArchive__b` before being deleted by `LogBatchPurger`
- I've also made a few tweaks to the index for `LogEntryArchive__b`, but I'm guessing we'll probably try a few other variations on what to use within the index
- I also included some placeholder metadata that will eventually be used to display `LogEntryArchive__b` records via a new LWC + custom tab

## Slack Plugin Changes & Log Retention Rules Plugin Changes
- The metadata for both plugins have been updated to use the latest changes to the plugin framework - there shouldn't be any functional changes


jongpie:
@jamessimone there are a few failing tests in this PR that I'm not too worried about for this PR - 1 is an issue for Flows that I'll fix in the `plugin-framework-overhaul` branch, and a few test failures are because I've only created the skeleton for a new LWC/tab to show `LogEntryArchive__b` (and I included some `System.assert(false)` calls as placeholders/reminders)

jamessimone:
Awesome. Incredible write-up — looking forward to reviewing!

nebula-logger/core/main/configuration/classes/LoggerParameter.cls
```diff
@@ -340,10 +340,35 @@ public class LoggerParameter {
...
+        }
+        return matchingParameters;
+    }
+
+    private static Map<String, LoggerParameter__mdt> loadRecords() {
+        Map<String, LoggerParameter__mdt> parameters = LoggerParameter__mdt.getAll().clone();
+        if (System.Test.isRunningTest() == true) {
+            parameters.clear();
+            for (String testContextParameterName : PARAMETERS_TO_LOAD_DURING_TESTS) {
+                if (LoggerParameter__mdt.getAll().containsKey(testContextParameterName) == true) {
```
jamessimone, line 358: shouldn't this be `parameters.containsKey(testContextParameterName) == true` ?

jamessimone: oh I see, you've cleared it above. seems like there could be a little optimization here as far as comparing to `parameters` prior to clearing for the test run so you don't have to keep referring to the map; on the other hand, I guess the `getAll()` map is cached, so maybe it's not a big deal

nebula-logger/core/main/plugin-framework/classes/LoggerPlugin.cls
```diff
@@ -0,0 +1,134 @@
...
+        }
+        matchingPluginConfigurationSorters.sort();
+        List<LoggerPlugin__mdt> matchingPluginConfigurations = new List<LoggerPlugin__mdt>();
+        for (PluginConfigurationSorter sorter : matchingPluginConfigurationSorters) {
+            matchingPluginConfigurations.add(sorter.pluginConfiguration);
+        }
+        return matchingPluginConfigurations;
+    }
+
+    public static List<LoggerPlugin__mdt> getSObjectHandlerPluginConfigurations() {
```
jamessimone, line 49: this method, and `getBatchPurgerPluginConfigurations()` are so similar that I wonder if there's room for a shared method that takes in the flow/apex `Schema.SObjectField` tokens and returns the sorted plugins based on that 🤔

nebula-logger/core/tests/log-management/classes/LogBatchPurger_Tests.cls
```diff
@@ -191,4 +198,132 @@ private class LogBatchPurger_Tests {
...
+
+        LogBatchPurger.BatchPurgerInput expectedInput = new LogBatchPurger.BatchPurgerInput(mockBatchableContext, instance.currentSObjectType);
+        System.assertEquals(false, ranPluginStart);
+        System.assertEquals(false, ranPluginExecute);
+        System.assertEquals(true, ranPluginFinish);
+        System.assertEquals(mockPluginConfiguration, pluginConfiguration);
+        System.assertEquals(expectedInput.sobjectType, pluginInput.sobjectType);
+    }
+
+    public class MockBatchableContext implements Database.BatchableContext {
```
jamessimone, line 288: nice!

nebula-logger/core/tests/log-management/classes/LoggerSObjectHandler_Tests.cls
```diff
@@ -154,46 +154,57 @@ private class LoggerSObjectHandler_Tests {
...
+        LoggerPlugin.setMock(plugin);
 
-        handler.execute();
+        Log__c log = new Log__c(LoggedBy__c = UserInfo.getUserId(), TransactionId__c = '6789');
+        insert log;
 
-        System.assertEquals(1, handler.executionCount);
-        System.assertEquals(1, handler.getPluginConfigurations().size());
-        System.assertEquals(0, handler.getExecutedApexPlugins().size());
+        String expectedProfileName = 'Some String';
```
jamessimone, line 205: It really is a shame that inner classes can't have static inner properties - how do you feel about promoting this to top-level constant in this class?

jamessimone (approved):

nebula-logger/core/tests/log-management/classes/LoggerSObjectHandler_Tests.cls
jongpie: Ooooh, yeah, I love the idea of promoting it to a top-level constant! I'll make that change. It's definitely a shame that inner classes can't have static stuff, I run into frustrations with that far too often.

nebula-logger/core/tests/log-management/classes/LogBatchPurger_Tests.cls
jongpie: Yeah, I generally don't reference `Database.BatchableContext` in my batchable classes, so in tests, I typically just pass `null` for the parameter in `start()` / `execute()` / `finish()`, but having a small mock implementation seemed useful in this case

nebula-logger/core/main/configuration/classes/LoggerParameter.cls
jongpie: Yeah, that's a good point - I don't think there will currently be much performance difference because right now, there's only a single `LoggerParameter__mdt` record that meets this use case, but I still like the idea of optimizing this a little more. I'll play around with a few options to see if I can optimize it a bit.

nebula-logger/core/tests/log-management/classes/LogBatchPurger_Tests.cls
jamessimone: Same, but it’s still really cool to see an actual usage

nebula-logger/core/main/plugin-framework/classes/LoggerPlugin.cls
jongpie: Hmmm, yeah, I like that idea - that would reduce long-term maintenance/changes to the `LoggerPlugin` class + makes it easier elsewhere in the codebase to do things dynamically, to use filter on a list of fields, etc. I think I'm going to go with something like this new method signature:

`public static List<LoggerPlugin__mdt> getFilteredPluginConfigurations(List<Schema.SObjectField> populatedFilterFields, Schema.SObjectField sortByField)`

And then to replace the method `getSObjectHandlerPluginConfigurations()`, the new method will be called like this to get the same results:

```java
LoggerPlugin.getFilteredPluginConfigurations(
  new List<Schema.SObjectField>{ Schema.LoggerPlugin__mdt.SObjectHandlerApexClass__c, Schema.LoggerPlugin__mdt.SObjectHandlerFlowName__c },
  Schema.LoggerPlugin__mdt.SObjectHandlerExecutionOrder__c
);
```

Similarly, `LoggerPlugin.getBatchPurgerPluginConfigurations()` (used in `LogBatchPurger`) will also be replaced with: 
```java
this.pluginConfigurations = LoggerPlugin.getFilteredPluginConfigurations(
  new List<Schema.SObjectField>{ Schema.LoggerPlugin__mdt.BatchPurgerApexClass__c, Schema.LoggerPlugin__mdt.BatchPurgerFlowName__c },
  Schema.LoggerPlugin__mdt.BatchPurgerExecutionOrder__c
);
```

I also like that this change then delegates the responsibility of deciding "which fields to use" within `LogBatchPurger` and `LoggerSObjectHandler` - I feel like they should handle that logic, `LoggerPlugin` class shouldn't know that much about the usage of the CMDT records.

jamessimone: 🎉🎉

PR #318, opened 2022-05-26
Added new method Logger.logDatabaseErrors() (plus overloads)

### LINKED ISSUE #295: Add helper methods for Database.XResult and Database.XResult lists to log on errors
### New Feature Summary

Currently, classes with multiple DML operations happening at once where `allAtOnce` is set to false have to manually iterate through their Database.SaveResult, Database.UpsertResult, Database.DeleteResult, Database.UndeleteResult, etc ... when deciding what to log.

I'd like to propose that Logger expose a helper method for _only logging errors_ such that:

```java
List<Object> dmlOperationsToVerify = new List<Object>();
dmlOperationsToVerify.addAll(Database.insert(someListToInsert, false));
dmlOperationsToVerify.addAll(Database.delete(someListToDelete, false));
Logger.logDMLErrors(dmlOperationsToVerify, LoggingLevel.WARN);
```

Not tied to the method name `logDMLErrors` by any means, but it would be nice for Logger to be able to do the list iteration - regardless of the underlying Database result type - and only log where `.isSuccess()` is false. It's ... highly unfortunate ... that there's no base class for all of the different database result classes, but Logger having a method like this would help to ease the pain associated with working with these DML operation lists.

jongpie:
This resolves #295 by adding a new method `Logger.logDatabaseErrors()`, including several overloads for using LogMessage/String, as well as the different database result classes. When this method is called, if the list of database results includes 1 or more errors (based on `isSuccess() == false`), then a new log entry is added, and the JSON of any unsuccessful database results is serialized/stored in the log entry.

The full list of method overloads is:

```java
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, LogMessage logMessage, List<Database.DeleteResult> deleteResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, String message, List<Database.DeleteResult> deleteResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, LogMessage logMessage, List<Database.MergeResult> mergeResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, String message, List<Database.MergeResult> mergeResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, LogMessage logMessage, List<Database.SaveResult> saveResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, String message, List<Database.SaveResult> saveResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, LogMessage logMessage, List<Database.UpsertResult> upsertResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, String message, List<Database.UpsertResult> upsertResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, LogMessage logMessage, List<Database.UndeleteResult> undeleteResults);
global static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, String message, List<Database.UndeleteResult> undeleteResults);
```

### DIFF
--- a/nebula-logger/core/main/logger-engine/classes/Logger.cls
+++ b/nebula-logger/core/main/logger-engine/classes/Logger.cls
@@ -2429,6 +2429,211 @@ global with sharing class Logger {
     }
 
     /**
+     * @description Creates a log entry for any results within the provided List DeleteResult> where isSuccess() == true
+     * @param  loggingLevel  The logging level to use for the log entry
+     * @param  logMessage    The instance of `LogMessage` to use to set the entry's message field
+     * @param  deleteResults The instance of `List<Database.DeleteResult>` to log
+     * @return               The instance of `LogEntryBuilder` was generated to log any errors, or `null` if there are no errors
+     */
+    public static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, LogMessage logMessage, List<Database.DeleteResult> deleteResults) {
+        if (isEnabled(loggingLevel) == false) {
+            return null;
+        } else {
+            return logDatabaseErrors(loggingLevel, logMessage.getMessage(), deleteResults);
+        }
+    }
+
+    /**
+     * @description Creates a log entry for any results within the provided List DeleteResult> where isSuccess() == true
+     * @param  loggingLevel  The logging level to use for the log entry
+     * @param  message       The string to use to set the entry's message field
+     * @param  deleteResults The instance of `List<Database.DeleteResult>` to log
+     * @return               The instance of `LogEntryBuilder` was generated to log any errors, or `null` if there are no errors
+     */
+    public static LogEntryEventBuilder logDatabaseErrors(LoggingLevel loggingLevel, String message, List<Database.DeleteResult> deleteResults) {
+        if (isEnabled(loggingLevel) == false) {
+            return null;
+        }
+
+        List<Database.DeleteResult> resultsToLog = new List<Database.DeleteResult>();
+        for (Database.DeleteResult deleteResult : deleteResults) {
+            if (deleteResult.isSuccess() == false) {
+                resultsToLog.add(deleteResult);
+            }
+        }
+
+        LogEntryEventBuilder logEntryEventBuilder;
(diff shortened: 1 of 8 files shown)


jongpie:
@jamessimone if you have some time this week/next week to review, let me know what you think of these changes. I was back & forth on if it made more sense to create a single log entry per list (which is what I've implemented in this PR), or if each unsuccessful result should have a separate, individual log entries - I figured that a separate log entry per result would consume a lot more data storage, so I went with creating 1 log entry per list of unsuccessful results.

nebula-logger/core/main/logger-engine/classes/Logger.cls
```diff
@@ -2428,6 +2428,211 @@ global with sharing class Logger {
         return finest().setMessage(message);
     }
 
+    /**
+     * @description Creates a log entry for any results within the provided List DeleteResult> where isSuccess() == true
+     * @param  loggingLevel  The logging level to use for the log entry
```
jamessimone, line 2433: != true, right?

jamessimone:
@jongpie just the one comment about the Apex Doc comments using `== true` instead of `!= true`, otherwise I totally agree with the approach and I think this all looks great!

jamessimone (approved):

nebula-logger/core/main/logger-engine/classes/Logger.cls
jongpie: Whooops! Yup, I'll fix these comments + regenerate the docs files. Thanks for catching this!

jongpie:
@jamessimone thanks so much for reviewing this! I'll fix those ApexDoc comments - I also realized I need a few more test methods to cover when a specified logging level is disabled + when no errors are found in the provided list of database results. I'll wrap up those couple of those changes, and then this should be good to merge! 🥳

jamessimone:
Aw yeah!

PR #333, opened 2022-06-15
Fix Slack Plugin URLs

dmgerow:
Current slack plugin links point to the wrong directory.


jongpie:
@dmgerow this is great, thanks for finding & fixing this!

jongpie (approved):

Issue #355, opened 2022-08-28
Add more fields to LogScenarioRule__mdt


jongpie:
Currently, the `LogScenarioRule__mdt` object provides some basic controls over Nebula Logger's behavior for a specified scenario. However, it does not have equivalent overrides for every feature currently controlled by `LoggerSettings__c`. Using log scenario rules, admins & devs should be able to override none/some/all feature flags that live on `LoggerSettings__c`. To support this, these new fields should be added to `LogScenarioRule__mdt`. There should also be a way to either expire a `LogScenarioRule__mdt` at a specified `Datetime`, or even better - have it only be active during a specified start & end datetime.

### Existing Custom Fields

-   `IsEnabled__c`
-   `NumberOfDaysToRetainLogs__c`
-   `Scenario__c`
-   `UserLoggingLevel__c`

### All Overrides for `LoggerSettings__c` Should Be Optional

Admins, developers (as well as ISVs) should be able to only specify the fields that they want to override on `LoggerSettings__c` - for a few reasons

-   **Logger Settings records can more easily be updated on-the-fly:** Admins & developers with the `LoggerAdmin` permission set can leverage the 'Logger Settings' tab/LWC to fine-tune logging features by creating/updating `LoggerSettings__c` records, instead of having to redeploy `LogScenarioRule__mdt` records.
-   **Custom Metadata Types don't provide profile & user-specific levels of configuration:** One of the best parts of `LoggerSettings__c` is that it's a custom hierarchy setting, allowing admins & developers to configure it at the organization, profile, and user levels. This makes it really easy to customize Nebula Logger differently for differents users - the `LogScenarioRule__mdt` custom metadata type does not have this option, so making its override fields optional gives admins & developers a way to still leverage the advantages of a custom hierarchy setting, while being able to override certain Nebula Logger feature flags.
-   **New fields may be added in the future:** - regression issues could be introduced in future releases if Nebula Logger assumed all fields are required on `LogScenarioRule__mdt`), and new fields are added, and existing records aren't updated to populate the new fields.

In order to make `Boolean` field optionals, the only viable option I can think of is to define the fields on `LogScenarioRule__mdt` as optional picklists with values of 'true' and 'false', and then parse the strings to `Boolean` values. It's... not my favorite, but would provide the desired functionality of "each override should be optional" & the implementation should be easy to implement/maintain.

### Field Naming Conventions:

[...]

PR #356, opened 2022-08-29
Created new LoggerScenario__c custom object

### LINKED ISSUE #349: Add new object LogScenario__c
### New Feature Summary

In the current implementation (<=4.7.8), using `Logger.setScenario(String)` populates the text field `Log__c.Scenario__c`, which is great for basic reporting & list view filtering. But if there were instead a new custom object `LogScenario__c` used (with a new lookup field `Log__c.LogScenario__c`), then orgs could leverage the log scenario records for additional features, such as assigning log scenarios to users/queues, adding sharing rules, etc. This has a downside of using more data storage, but it would be a more robust way to manage scenarios.

jongpie:
Closes #349 by deprecating the field `Log__c.Scenario__c` and replacing it with a new optional lookup `Log__c.TransactionScenario__c` that leverages the new custom object `LoggerScenario__c `. The `LoggerScenario__c ` object does not currently have any automation built-in - this PR is focused on just getting the object introduced, more features will be added later.

- Added `LoggerScenario__c ` object with a uniqued external ID field `LoggerScenario__c .Unique__c` - this is set based on the value `LogEntryEvent__e.Scenario__c` (which can be set using `Logger.setScenario(String)`
- Added new "Logger Scenarios" tab to the Logger Console app, and rearranged the ordering of the included tabs
  
  image

- Added skeleton trigger handler class `LoggerScenarioHandler` - there's currently no object-specific logic implemented, but this provides an entry point for plugins. Additional logic will probably also be incorporated as part of #355 
- Added flexipage, page layout, compact layout, etc for the new object. I've arranged the UI to mimic the UI of the `Log__c` object's flexipage/layout

  image

- Added an Apex script `./scripts/data/migrate-log-scenario-field-to-logger-scenario-object.apex` that migrates existing data from the text field `Log__c.Scenario__c` to the new object & lookup `Log__c.TransactionScenario__c`. 
  - Existing orgs can run the script after upgrading to update existing records - but depending on how many `Log__c` records need to be updated, the script may need to be executed multiple times. It's not ideal, but this approach is easier that deploying a new, temporary batch job to do it in bulk.
- Updated `LoggerAdmin`, LoggerLogViewer`, and `LoggerEndUser` permission sets to have access to the new `LoggerScenario__c` object
- Updated `LoggerEndUser` permission set to have read-only access to the tagging & scenario custom objects


jongpie:
@jamessimone after we chatted about this earlier today (thanks again!), I'm going to rename a few things in this PR & PR #357
- In this PR, I'm going to rename the new custom object `LogScenario__c` to `LoggerScenario__c` so it's a little bit more generic sounding, and lines up with the method name `Logger.setScenario()`. I'll probably keep the name as is for the new lookup field `Log__c.LogScenario__c` (since `Scenario__c` already exists as a text field)
- In #357, I'm thinking of also renaming `LogScenarioRule__mdt` to be called  `LoggerScenarioRule__mdt` so it's consistent with `LoggerScenario__c`
  - I'll also try to incorporate your idea of having individual  `LogEntry__c` fields tied to the new scenario custom object, as well as the possible new method `Logger.endScenario()`

Other than the object rename, I'm not planning to make any other changes in this PR - I'm aiming to have the rename completed in the next day or so

jamessimone:
@jongpie sounds good! I will take a look later today 🎉

nebula-logger/core/main/log-management/classes/LogHandler.cls
```diff
@@ -27,6 +28,7 @@ public without sharing class LogHandler extends LoggerSObjectHandler {
 
     protected override void executeBeforeInsert(List<SObject> triggerNew) {
         this.logs = (List<Log__c>) triggerNew;
+        this.logScenariosById = queryLogScenarios(this.logs);
```
jamessimone, line 31: missing the `this` prefix here and in the below usage of this method, not sure if that was intentional or not

nebula-logger/core/main/log-management/objects/Log__c/fields/ApiVersion__c.field-meta.xml
```diff
@@ -16,6 +16,11 @@
     <valueSet>
         <valueSetDefinition>
             <sorted>false</sorted>
+            <value>
+                <fullName>v57.0</fullName>
+                <default>false</default>
+                <label>v57.0 - Spring &apos;23 Release</label>
```
jamessimone, line 22: coming right up!

nebula-logger/core/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -836,7 +908,12 @@ private class LogEntryEventHandler_Tests {
         System.assertEquals(logEntryEvent.ParentLogTransactionId__c, log.ParentLogTransactionId__c, 'log.ParentLogTransactionId__c was not properly set');
         System.assertEquals(logEntryEvent.ProfileId__c, log.ProfileId__c, 'log.ProfileId__c was not properly set');
         System.assertEquals(logEntryEvent.ProfileName__c, log.ProfileName__c, 'log.ProfileName__c was not properly set');
-        System.assertEquals(logEntryEvent.Scenario__c, log.Scenario__c, 'log.Scenario__c was not properly set');
+        System.assertEquals(
+            logEntryEvent.Scenario__c?.left(Schema.LogScenario__c.Name.getDescribe().getLength()),
```
jamessimone, line 912: what's the mechanism here that's setting the Scenario here? I didn't see anything in setup creating a scenario record, and the assert makes it seem like there could be truncation at play here, as well?

jamessimone: Oh maybe it's just not part of the diff because I see you were previously asserting on the Scenario being set ...

jamessimone (approved):
@jongpie LGTM! Just a couple of comments here and there but everything looks great!

jongpie:
@jamessimone thanks for reviewing! I'll finish the object rename & incorporating your comments, and then I'll publish this release today or tomorrow!

PR #357, opened 2022-09-01
More controls for scenario-based logging

### LINKED ISSUE #347: Is it possible to configure Logger to not log for specific Apex classes/triggers?
Or would I need to write custom code to disable the Logger wherever needed?

### LINKED ISSUE #355: Add more fields to LogScenarioRule__mdt

### LINKED ISSUE #360: NullPointer Exception when addTag method is called upon the return Object of Logger.logDatabaseErrors()

Unlocked Package

v.4.8.0

### New Bug Summary

addTag() method throws a NullPointer Exception when is called from the LogEntryEventBuilder return Object of logDatabaseErrors().

### LINKED ISSUE #363: Log scenario migration script bug
There's a bug on line 16 of the transaction scenario migration script.

`System.SObjectException: SObject row was retrieved via SOQL without querying the requested field: Log__c.TransactionScenario__r`

Need to add this field to the select.

[https://github.com/jongpie/NebulaLogger/blob/5f59b7a9d04e1e7f0d649e9712f82a74961db86d/scripts/data/migrate-log-scenario-field-to-logger-scenario-object.apex#L16](url)

jongpie:
### Core package changes
This PR is complements some of the changes released in #356 (`v4.8.1`), and closes #355 by further enhancing scenario-based logging
- Fixed #360 by changing Logger.logDatabaseErrors() overloads to always return a non-null instance of LogEntryEventBuilder
- Updated backend code for LWC `relatedLogEntries` to remove a poorly-written & non-functioning Apex method that was trying determine the correct tab icon for the `LogEntry__c` object. Hardcoding the icon name into the LWC is simpler, uses less resources, and actually functions 😅 
- Added new method `Logger.endScenario(String scenario)`. This provides a way for Apex developers to explicitly end a scenario, which then automatically rolls back to the previous scenario when applicable (and re-applies the previous scenario rule, if one exists)
- Added new fields `LogEntryEvent__e.EntryScenario__c` and `LogEntry__c.EntryScenario__c` (used in addition to `LogEntryEvent__e.TransactionScenario__c` and `Log__c.TransactionScenario__c`). If multiple scenarios are specified during a transaction, the `EntryScenario__c` fields will indicate the specific scenario that was active when an entry was created. 
- Deprecated `LogScenarioRule__mdt` object, replaced it with new object `LoggerScenarioRule__mdt`. This, combined with `Logger.endScenario()`, closes #347 
  - Using `LoggerScenarioRule__mdt`, some/all fields on `LoggerSettings__c` can be optionally overridden, allowing more control for different modules/groupings of code with an org
  - Using the optional `StartTime__c` and `EndTime__c` datetime fields, you can control when a `LoggerScenarioRule__mdt` should be used. This can be useful in situations where you need to temporarily override some `LoggerSettings__c` for your scenario
  - `LogScenarioRule__mdt` Deprecated Object:

    image

  - `LoggerScenarioRule__mdt` Revamped Object:

    image

### Slack plugin package changes

- Small cosmetic change in the messages posted to Slack - the fields `LogEntry__c.StackTrace__c` and `LogEntry__c.ExceptionStackTrace__c` have been updated to be formatted as code blocks, instead of inline code snippets. This works better in Slack's UI when the fields contain line breaks.
- Created new package version `v1.5.0` (making a somewhat arbitrary jump from `v0.10.0`, the old version numbers didn't align with the fact that the Slack plugin is production-ready)

### Other changes
- Fixed #363 by correcting some references in the migration script 
- Removed build numbers from all package versions listed in `sfdx-project.json` - the updated version numbers are now consistent with the version number tags published in GitHub


jongpie:
@jamessimone I think this PR is finally ready for review! Let me know your thoughts whenever you've had a chance to take a look.

nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls
```diff
@@ -66,28 +66,51 @@ public without sharing class LogEntryEventHandler extends LoggerSObjectHandler {
         List<LogEntryEvent__e> logEntryEventsToSave = new List<LogEntryEvent__e>();
         for (LogEntryEvent__e logEntryEvent : newLogEntryEvents) {
             User loggingUser = new User(Id = logEntryEvent.LoggedById__c, ProfileId = logEntryEvent.ProfileId__c);
-            LoggerSettings__c loggingUserSettings = Logger.getUserSettings(loggingUser);
-            if (loggingUserSettings.DefaultPlatformEventStorageLocation__c == DEFAULT_STORAGE_LOCATION_NAME) {
+            String platformEventStorageLocation = Logger.getUserSettings(loggingUser).DefaultPlatformEventStorageLocation__c;
+            if (logEntryEvent.TransactionScenario__c != null && LoggerScenarioRule.getAll().containsKey(logEntryEvent.TransactionScenario__c) == true) {
+                LoggerScenarioRule__mdt scenarioRule = LoggerScenarioRule.getInstance(logEntryEvent.TransactionScenario__c);
+                if (scenarioRule.IsPlatformEventStorageLocationEnabled__c == String.valueOf(true)) {
```
jamessimone, line 72: Might want to store `String.valueOf(true)` outside of the for loop

nebula-logger/core/main/log-management/lwc/relatedLogEntries/relatedLogEntries.html
```diff
@@ -5,7 +5,7 @@
 
 <template>
     <template if:true={showComponent}>
-        <lightning-card title={title} icon-name={queryResult.tabIcon} class="log-entry-card">
+        <lightning-card title={title} icon-name="custom:custom83" class="log-entry-card">
```
jamessimone, line 8: nice

nebula-logger/core/main/logger-engine/classes/Logger.cls
```diff
@@ -2686,36 +2635,80 @@ global with sharing class Logger {
...
-        for (LogScenarioRule__mdt scenarioRule : LogScenarioRule__mdt.getAll().values()) {
-            if (scenarioRule.IsEnabled__c == true && scenarioRule.Scenario__c == scenario) {
-                matchingScenarioRule = scenarioRule;
-                break;
+        currentEntryScenario = scenario;
+        orderedScenarios.add(scenario);
+        LoggerScenarioRule__mdt matchingScenarioRule = LoggerScenarioRule.getInstance(scenario);
+        if (matchingScenarioRule != null) {
+            LoggerSettings__c userSettings = getUserSettings();
+            if (String.isNotBlank(matchingScenarioRule.IsLoggerEnabled__c) == true) {
```
jamessimone, line 2660: I think that we did chat about this earlier, but can you refresh my memory on the thought process behind implementing these as picklists? Also, does it make sense to have one global value set for them if they're all going to be true/false?

nebula-logger/core/main/logger-engine/classes/Logger.cls
```diff
@@ -2933,19 +2928,6 @@ global with sharing class Logger {
...
-        // if (String.isBlank(scenarioRule.Scenario__c) == true) {
-        //     throw new IllegalArgumentException('Scenario__c is required on `LogScenarioRule__mdt: \n' + JSON.serializePretty(scenarioRule));
-        // }
-
-        MOCK_SCENARIO_TO_SCENARIO_RULE.put(scenarioRule.Scenario__c, scenarioRule);
-    }
-
     private static SaveMethod getSaveMethod(String saveMethodName) {
         try {
             return SaveMethod.valueOf(saveMethodName);
```
jamessimone, line 2933: I did some profiling on `Enum.valueOf(value)` and found performance pretty wanting for repeated calls to it. Might be a TODO for later but if `getSaveMethod` is being called more than once per transaction it unfortunately might be more efficient to cache a map of string to the individual save methods yourself

nebula-logger/core/tests/log-management/classes/LogEntryEventHandler_Tests.cls
```diff
@@ -89,14 +89,55 @@ private class LogEntryEventHandler_Tests {
         System.assertEquals(0, countOfLogEntries);
     }
 
+    @IsTest
+    static void it_should_not_create_log_or_log_entry_data_when_platform_event_storage_location_is_null_in_logger_scenario_rule() {
```
jamessimone, line 93: what's the use-case for not having a storage location on the scenario rule? wouldn't that be a waste of CPU time to even create a log entry record in memory at all?

nebula-logger/core/tests/logger-engine/classes/Logger_Tests.cls
```diff
@@ -166,27 +166,133 @@ private class Logger_Tests {
...
+        LogEntryEvent__e firstLogEntryEvent = publishedLogEntryEvents.get(0);
+        System.assertEquals(firstMockScenario, firstLogEntryEvent.EntryScenario__c);
+        System.assertEquals(firstMockScenario, firstLogEntryEvent.TransactionScenario__c);
+        LogEntryEvent__e secondLogEntryEvent = publishedLogEntryEvents.get(1);
+        System.assertEquals(secondMockScenario, secondLogEntryEvent.EntryScenario__c);
+        System.assertEquals(firstMockScenario, secondLogEntryEvent.TransactionScenario__c);
+    }
+
+    @IsTest
+    static void it_should_end_scenario_when_no_scenario_was_previously_set() {
```
jamessimone, line 229: ha, love this one

jamessimone (approved):
@jongpie looks good to me! I left a few comments here and there. Thanks in advance for bringing me back up to speed on the decision to use picklists for the `IsEnabled...__c` properties, and then the only other thing I was looking for clarification on is the question I left in `LogEntryEventHandler_Tests` about when the platform even storage location isn't specified on the logger scenario rule

nebula-logger/core/main/logger-engine/classes/Logger.cls
jongpie: Ooooh, that's a great idea - I'll add a TODO to get that changed in another branch I'm working on for some optimizations.

nebula-logger/core/main/logger-engine/classes/Logger.cls
jongpie: I want all of these fields (that override a field on `LoggerSettings__c`) to be optional overrides - that way, if you're creating your own `LoggerScenarioRule__mdt`, you can choose to only override a subset of `LoggerSettings__c` fields. You may still want `LoggerSettings__c` to control _most_ of the settings. But since `Boolean` fields can't be null in Salesforce, the easiest (although a bit gross) way I've found to make it optional is to make it an optional picklist field, and use a `null` check in the code to see if the field should be overridden.

And I just double checked in my scratch org - for some mysterious reason, you can't use a global value set for picklists on a custom metadata type object. It seems like a feature gap on the platform unfortunately.

image

nebula-logger/core/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jongpie: Several orgs use it for publishing platform events (and they have external systems subscribing to the events) without using any data storage in the org. It's a bit of an edge case, but it seems like bigger orgs with a lot of logging will sometimes disable storage on platform for certain users, and have the platform events stored off-platform instead.

nebula-logger/core/tests/logger-engine/classes/Logger_Tests.cls
jongpie: Some of these method names were fun to write

nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls
jongpie: Yeah, that sounds good, that should be a quick change.

nebula-logger/core/main/log-management/lwc/relatedLogEntries/relatedLogEntries.html
jongpie: The old code for determining the tab icon was so clunky, and apparently stopped working at some point, so this seemed like one of those rare exceptions where hardcoding a string was better 😅

nebula-logger/core/main/logger-engine/classes/Logger.cls
jongpie: Nevermind, I'm just going to implement this now 😄 I wrote the TODO in the class locally, but the actual code changes are about as long as the TODO, so I'm just going include the optimization with this PR. The first time that the method is called, it will now cache the enum names & values in a new private map constant `SAVE_METHOD_NAME_TO_SAVE_METHOD`:

```java
private static SaveMethod getSaveMethod(String saveMethodName) {
    if (SAVE_METHOD_NAME_TO_SAVE_METHOD.isEmpty() == true) {
        for (SaveMethod method : SaveMethod.values()) {
            SAVE_METHOD_NAME_TO_SAVE_METHOD.put(method.name(), method);
        }
    }

    return SAVE_METHOD_NAME_TO_SAVE_METHOD(saveMethodName);
}
```

nebula-logger/core/main/logger-engine/classes/Logger.cls
jamessimone: Ah yes, I remember now. I knew we had talked about this! Thanks for clarifying.

nebula-logger/core/tests/log-management/classes/LogEntryEventHandler_Tests.cls
jamessimone: Right, right - ok, that makes sense!

jongpie:
Thanks @jamessimone for reviewing! I'm incorporating some of your feedback now - let me know if you have any follow up questions about my responses on the boolean-ish picklist fields or the `LogEntryEventHandler_Tests` platform event storage location use case.

PR #392, opened 2022-10-11
Flow Execution Error Event Handling

### LINKED ISSUE #372: Add logging for FlowExecutionErrorEvent in Async Failures Addition plugin
Salesforce has a platform event object called [`FlowExecutionErrorEvent`](https://developer.salesforce.com/docs/atlas.en-us.platform_events.meta/platform_events/sforce_api_objects_flowexecutionerrorevent.htm) that generates events for errors in screen flows (autolaunched flows are not supported unfortunately). It seems conceptually similar to Apex's `BatchApexErrorEvent`, so this would be a nice fit for the [Async Failures Additions plugin](https://github.com/jongpie/NebulaLogger/tree/main/nebula-logger/plugins/async-failure-additions).

I'm not sure exactly where some of the fields on `FlowExecutionErrorEvent` would map to in Nebula Logger, but hopefully some of the existing Flow-specific fields on `LogEntryEvent__e` and `LogEntry__c` can be reused.

jamessimone:
Fixes #372 by adding support for FlowExecutionErrorEvent via platform event-triggered flow

### DIFF
--- a/nebula-logger/plugins/async-failure-additions/plugin/customMetadata/LoggerParameter.FlowExecutionErrorEventHandled.md-meta.xml
+++ b/nebula-logger/plugins/async-failure-additions/plugin/customMetadata/LoggerParameter.FlowExecutionErrorEventHandled.md-meta.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<CustomMetadata
+    xmlns="http://soap.sforce.com/2006/04/metadata"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+>
+    <label>Is FlowExecutionErrorEvent Handled?</label>
+    <protected>true</protected>
+    <values>
+        <field>Description__c</field>
+        <value xsi:type="xsd:string">1 if enabled, 0 if disabled. True values create Nebula Logger entries for unhandled flow errors.</value>
+    </values>
+    <values>
+        <field>Value__c</field>
+        <value xsi:type="xsd:string">0</value>
+    </values>
+</CustomMetadata>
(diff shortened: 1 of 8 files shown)


nebula-logger/plugins/async-failure-additions/plugin/customMetadata/LoggerParameter.FlowExecutionErrorEventHandled.md-meta.xml
```diff
@@ -0,0 +1,17 @@
...
+>
+    <label>Is FlowExecutionErrorEvent Handled?</label>
+    <protected>true</protected>
+    <values>
+        <field>Description__c</field>
+        <value xsi:type="xsd:string">1 if enabled, 0 if disabled. True values create Nebula Logger entries for unhandled flow errors.</value>
+    </values>
+    <values>
+        <field>Value__c</field>
+        <value xsi:type="xsd:string">0</value>
```
jongpie, line 15: Shouldn't this be a value of 'true' or 'false' (instead of '1' or '0')? It looks like the code is treating it as a boolean.

nebula-logger/plugins/async-failure-additions/plugin/flows/Nebula_Logger_Flow_Execution_Error_Event_Handling.flow-meta.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
jongpie, line 1: Just for consistency with other metadata filenames, could this flow be renamed to something like `LoggerFlowExecutionErrorEventHandling`?

jongpie (commented):
@jamessimone this looks really awesome! It looks like you only had to do minimal field overrides to get this working, I really like it. I added 2 small comments, let me know what you think.

nebula-logger/plugins/async-failure-additions/plugin/customMetadata/LoggerParameter.FlowExecutionErrorEventHandled.md-meta.xml
jamessimone: @jongpie that's what I initially thought as well, and I swear I had an error using those string values, but now everything seems to be working as intended so I will swap to true/false.

jamessimone:
@jongpie just updated based on your feedback!

PR #394, opened 2022-10-18
Query Optimizations & More Configurable Controls

### LINKED ISSUE #311: Optimizations for Synchronous Context
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/303

<sup>Originally posted by **LawrenceLoz** April 28, 2022</sup>
Hi there,

Firstly big kudos to author/authors and contributors on this project - a really rich and intuitive set of features and some great configuration flexibility.

Running a few experiments and looking at debug logs, I've noticed that when the Logger object class is called several queries are run within the transaction which contribute to the limit of 100 SOQL statements - in my tests this is 4 queries and I think would be 5 for logs being made for community users. Whilst this isn't an especially high, in an ideal world given how widely a logging framework should be used it would be great to reduce the utilisation and improve synchronous performance where possible.

There are a couple of areas resulting in queries being run, and I wonder whether it's viable to refactor to avoid these:

1. Line 2556 in Logger initialises LogEntryEventHandler: `LogEntryEventHandler().executeBeforeInsert(logEntryEvents);` As part of this, TAG_ASSIGNMENT_RULES variable is set from the results of a CMDT query, which includes flitering on SObjectField__r.QualifiedApiName (the bit which makes the CMDT query contribute to limits like a normal SOQL object query). This variable isn't used until after PEs are created so feels like this should be called only at that stage
2. There are methods queryOrganization(), queryUser(), queryAuthSession() and queryNetwork() in the LogEntryEventBuilder, which all query for details to populate fields on the event. Could it make sense instead to populate just the user id on the event and avoid checking for this extra information until LogEntryEventHandler is processing the events? The downside of this approach is that events would be missing the extra info so an external system consuming these wouldn't have the detail, so maybe the choice of whether these queries are made before or after events should be configurable in settings

I'm happy to create issues and help out with PRs suggesting changes for some or all of these areas if you agree there's a better approach? Or let me know if I'm missing something about how any of these can be avoided today?

Also a heads up I'm hoping to touch on Nebula Logger as part of a talk I'm co-presenting at London's Calling soon (https://www.londonscalling.net/sessions/session-apex-frameworks-for-architects/). Let me know if you'd like more insight into how this would be included (or if for any reason you'd prefer Nebula Logger not to be included!)

Again, thanks for the great work - I've developed a similarly architected logging framework myself and familiar with many of the challenges you've addressed really well here!</div>

Acceptance Criteria:
- [x] **Lazy-load query for `LogEntryEventHandler.TAG_ASSIGNMENT_RULES`** - Included in PR #307
- [x] Cache the call to `UserInfo.getSessionId()`
- [x] Add a configuration option to asynchronously run `queryAuthSession()`
- [x] Add a configuration option to asynchronously run `queryNetwork()`
- [x] Add a configuration option to asynchronously run `queryOrganization()`
- [x] Add a configuration option to asynchronously run `queryUser()`
- [x] Add a configuration option to skip running `LogEntryEventBuilder.parseStackTrace()`
- [x] Consider using platform cache to cache queries. Some queries would make sense to cache within the org partition; other queries (such as queries about the current user) should be cached within the session partition.
  - Since not all orgs have platform cache storage, this would need to be optional (probably controlled via a new `LoggerParameter__mdt` record?). If org/session cache isn't available (or disabled), then Nebula Logger should continue to use the existing transaction cache 

Next Steps:

- [x] In order to make the queries reusable (in both the synchronous and asynchronous contexts), all of the queries will need to be centralized into new class(es). This was completed in v4.7.3 (PR #314)
- [x] Need to consider the availability of platform cache in Professional and Essentials editions (and confirm availability for Enterprise & Unlimited)

### LINKED ISSUE #372: Add logging for FlowExecutionErrorEvent in Async Failures Addition plugin

### LINKED ISSUE #389: Leverage Platform Cache for API Release Version & Number
### New Feature Summary

We are using an old version (v4.6.16), but I noticed that in the LogEntryEventHandler class there is a SOQL query executed that queries the Log records created today or in the last 4 hours to retrieve the API Release Version and Number. This query is reasonably efficient. It uses the CreatedDate index, which is good. However, there is still a cost for this query and it may scan hundreds of thousands of rows with our volume. Also note that this SOQL query executes for every invocation of this event handler. There is also functionality to make a web callout about every 4 hours to the _status_ API to refresh the API release version & number.

This is likely only an issue for implementations with very high volume logging. We are currently seeing about 2M records / day and this may grow. We are looking for opportunities to reduce this now, but we are still bringing additional users on-board. Even still, it would be nice to have options to disable functionality that is "nice-to-have" in favor of adding scale & performance for high-volume implementations.

I think we could address this by adding 2 flags to the Nebula configuration.

A flag which will use platform cache to store the API Release Version and Number, rather than reading it from existing records via a SOQL query for every invocation of the LogEntryEventHandler.

We should have the option to disable the API release version & number capabilities where it does not make the web callout to the status API or attempt to store/read the values from Log records or platform cache.

### LINKED ISSUE #390: Flags to control potentially "expensive" functionality for high-volume implementations
### New Feature Summary

This issue kind of builds on top of Issue #389.

We are running an older version (v4.6.16), so some of this may be reflected in newer versions.

We also have a very high-volume implementation. We are currently seeing ~2M log records / day, and are still in the process of bringing more users on-board. We are looking to turn our log level down, but we still may have >1M logs / day.

At this volume, even "simple" queries that use an index may be expensive. Recently Salesforce Support raised a concern with the LogEntryEventHandler class saying that it was consuming 50% of DB CPU. That class only has a few SOQL queries and most are using an index or should not scan a large number of records. The issue is that this class is being invoked so many times each day that even small or simple queries are using a lot of DB CPU in the aggregate.

I spoke about this specific class in issue #389. What I want from this issue is to identify additional functionality in Nebula that might be "nice-to-have" and could be disabled to support high-volume implementations of Nebula. I think another example may be in LogEntryHandler where it is querying the record(s) associated with the Log Entry so that the record name can be put on the Log Entry record. While this is "nice-to-have", it is expensive and we would gladly forgo it in order to improve performance/scalability.

### LINKED ISSUE #393: LogManagementDataSelector_Tests.it_returns_profiles_for_specified_profile_ids unit test method is failing when some of user profile names begin with lower-case letter

Unlocked Package

v4.8.3

### New Bug Summary

`LogManagementDataSelector_Tests.it_returns_profiles_for_specified_profile_ids` unit test method is failing when some of user profile names begin with lower-case letter. 
E.g.:  The test will fail if there are three profiles named 'A', 'b', and 'C'.

This is because `SOQL ORDER BY` sorting is case-insensitive, and `List.sort()` method - is case-sensitive.

A solution is to sort both the expected and the returned lists in the same way.

jongpie:
# Core Package Changes

## Synchronous Context Optimizations
Several enhancements have been made to reduce the overhead that Nebula Logger consumes when logging - these enhancements help with some platform limits, such as query limits, CPU time limits, etc. (closes #311). Many thanks again to @LawrenceLoz for doing some stellar analysis on the codebase to identify some of these optimizations.
- Added new `LoggerParameter__mdt` record `EnableStackTraceParsing` to disable stack trace parsing org-wide. Although parsing stack traces can help provide contextual info for reporting, the code consumes CPU time in Apex - orgs can now choose if they want to disable this functionality to improve Nebula Logger's performance.
- Added new `LoggerParameter__mdt` records to enable/disable if `LogEntryEventBuilder` queries run synchronously (default behavior). When these records are disabled (`Value__c = 'false'`), the corresponding queries run asynchronously after platform events are published. This means that the corresponding fields aren't set on `LogEntryEvent__e`, but are still populated on `Log__c` and `LogEntry__c`. Four (4) new custom metadata records are included:
  1. `LoggerParameter.QueryAuthSessionDataSynchronously`
  2. `LoggerParameter.QueryNetworkDataSynchronously`
  3. `LoggerParameter.QueryOrganizationDataSynchronously`
  4. `LoggerParameter.QueryUserDataSynchronously`
- Updated `Logger` class to internally cached the value of `UserInfo.getSessionId()`. I had originally (erroneously) assumed that the platform would handle caching this, but repeated calls to the method seem to consume additional CPU time - the value is now cached in a private constant.

## Caching System Improvements
I've finally started leveraging [Platform Cache](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_cache_namespace_overview.htm) to supplement that existing transaction caching (partially closes #311 and closes #389)
- Added platform cache partion `LoggerCache` that comes with 3MB allocated, based on the 3MB that Salesforce provides for security-approved managed packages. Sadly, Nebula Logger has not yet gone through this process yet, but still seemed like a sensible default size to use.
	- 1MB allocated to the org partition
	- 2MB allocated to the session partition
- Updated Apex class `LoggerCache` to now support 3 types of caching:
	1. `LoggerCache.getTransactionCache()`. This is the existing caching system that's been used for a few releases. It caches values in a transaction using a `private static final Map<String, Object>` constant. This is still used internally throughout `LoggerCache` to supplement some limitations with platform cache, such as not being able to cache `null` values
	2. `LoggerCache.getOrganizationCache()`. This leverages the org partition of Platform Cache
	3. `LoggerCache.getSessionCache()`. This leverages the session partition of Platform Cache
- Updated all existing caching in Nebula Logger to now leverages platform caching by default. 
- Added a new `LoggerParameter__mdt` record 'UsePlatformCache' to control if Nebula Logger leverages platform cache. When 'UsePlatformCache' is disabled (Value__c = 'false'), Nebula Logger will only use `Logger.getTransactionCache()` (which is existing behavior before platform cache was introduced).

## Large Data Volume (LDV) Optimizations
I've included some query-related enhancements to address some of the performance problems that @rmccu brought up for orgs with large data volumes (2+ millions `Log__c` records created daily)
- Added new `LoggerParameter__mdt` records to completely disable certain queries executed by Nebula Logger (closes #390). Some of the queries used by Nebula Logger can perform poorly in orgs with large data volumes - the new records provide a way for orgs to completely disable certain queries, with the tradeoff that the corresponding fields will be `null`. Seven (7) queries can now be fully disabled - there are still some other queries throughout Nebula Logger that don't have this option, but these should be the most critical queries.
  1. `LoggerParameter.QueryApexClassData` - when disabled, the `ApexClass` object is not queried, and the relevant fields on `LogEntry__c` are null
  2. `LoggerParameter.QueryAuthSessionData` - when disabled, the `AuthSession` object is not queried, and the relevant fields on `LogEntry__c` are null
  3. `LoggerParameter.QueryFlowDefinitionViewData` - when disabled, the `FlowDefinitionView` object is not queried, and the relevant fields on `LogEntry__c` are null
  4. `LoggerParameter.QueryNetworkData` - when disabled, the `Network` object is not queried, and the relevant fields on `LogEntry__c` are null. Note: this only applies to orgs that use an Experience Cloud site (stored in the `Network` object)
  5. `LoggerParameter.QueryOrganizationData` - when disabled, the `Organzation` object is not queried, and the relevant fields on `LogEntry__c` are null
  6. `LoggerParameter.QueryRelatedRecordData` - when disabled, the dynamic queries used to query record names (based on the field `LogEntry__c.RecordId__c`) are not executed, the field `LogEntry__c.RecordName__c` is null
  7. `LoggerParameter.QueryUserData` - when disabled, the `User` object is not queried, and the relevant fields on `LogEntry__c` are null

## Bugfix for `LogEntryEvent__e.TimestampString__c`
This was an unreported bug that I stumbled across where ` LogEntryEvent__e.TimestampString__c` could be out of sync with `LogEntryEvent__e.Timestamp__c`, resulting in inaccurate data being stored in `LogEntry__c.Timestamp__c`. I've fixed the bug & added some additional tests to ensure that this continues to behave as expected.

## Data Model Changes
- Added new fields `LogEntryEvent__e.RequestId__c` and `Log__c.RequestId__c` to store the value of `System.Request.getCurrent().getRequestId()`. In versions v3.1.0 through v4.7.2, this value was used for `Log__c.TransactionId__c`, but I reverted to using a more reliable [custom `UUID`](https://github.com/jongpie/ApexUuid) in v4.7.3 (the original solution). The new `RequestId__c` fields provides a way to still report on this value, which can be helpful for tying Nebula Logger's data back to platform data & logs.

## Test Improvements
- Fixed #393 by removing an unnecessary & problematic call to `List.sort()` on the expected query results
- Added new `ApexTestSuite` LoggerExtraTests (only deployed to/used by the pipeline for integration tests)
- Added more integrations tests, primarily for Experience Site (Community/Network/Portal/too-many-names-ahhh) testing around `Network` querying & setting of related fields
- Make `LoggerEngineDataSelector` and `LogManagementDataSelector` classes & methods `virtual` & implemented test implementations in the respective test classes `LoggerEngineDataSelector_Tests` and `LogManagementDataSelector_Tests` to provide a cleaner way to test query executions within Nebula Logger's codebase. 
 - The test implementations currently usedin test classes still execute all of the related queries, so the queries aren't returning mock values (when possible) just yet. This PR has most of the foundational changes need to support full mocking, but I'm planning to revisit some of the tests throughout the codebase in a future release to implement mock return values when possible. That will make it easier to test queried data in general, and should also give the tests classes another nice boost in test speed (in addition to some other substantial test speed improvements made in v4.7.1).
 
 ## Experience Site Test Metadata Improvements
 These changes only impact the pipeline to help validate logic related to orgs that use Nebula Logger for Experience Cloud sites (aka Community Cloud) - these changes don't have any direct functional changes to Nebula Logger
 - Added high-volume traffic Community profile & related metadata changes for pipeline testing purposes
 - Added more Apex tests for Experience Cloud scenarios in the `extra-tests` folder. These tests are only used in the pipeline - they are not included in the core package.
 

# Async Failure Additions Plugin Package Changes
Created v1.0.2 of the Async Failure Additions plugin
- @jamessimone implemented an incredible, new feature for [this plugin](https://github.com/jongpie/NebulaLogger/tree/v4.8.3/nebula-logger/plugins/async-failure-additions) to automatically log Flow errors that occur in Screen Flows, using the platform event [`FlowExecutionErrorEvent`](https://developer.salesforce.com/docs/atlas.en-us.platform_events.meta/platform_events/sforce_api_objects_flowexecutionerrorevent.htm) (closes #372 - implemented in #392)

# Slack Plugin Package Changes
Created v1.5.1 of the Slack plugin
- Updated the formatting of the field 'Last Log Entry Message' to handle line breaks


jongpie:
@jamessimone I think this is finally ready for review! There's one issue with some of the managed package metadata that I'm trying to fix, but I'm hoping that will be some minimal changes to get it working - everything else should be ready.

jamessimone:
OK great! I'll try to take a look later today

nebula-logger/core/main/configuration/classes/LoggerCache.cls
```diff
@@ -0,0 +1,230 @@
...
+         *              using the specified `String` key
+         * @param  key   The `String` key to add to the session cache
+         * @param  value The `Object` value to cache for the specified key
+         */
+        public void put(String key, Object value) {
+            getTransactionCache().put(key, value);
+
+            // Platform cache does not support storing null values
+            if (LoggerParameter.USE_PLATFORM_CACHE == true && value != null && SESSION_CACHE_PARTITION.isAvailable() == true) {
+                SESSION_CACHE_PARTITION.put(key, value, SESSION_CACHE_TTL_SECONDS, CACHE_VISIBILITY, CACHE_IS_IMMUTABLE);
```
jamessimone, line 164: session cache supports nulls?

nebula-logger/core/main/configuration/classes/LoggerCache.cls
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /**
+         * @description Returns the cached value for the specified key, or `null` if
+         *              the specified key does not exist in the transaction cache
+         * @param  key  The `String` key to check for within the transaction cache
+         * @return      The cached value, or null if no cached value is found for the specified key
+         */
+        public Object get(String key) {
+            if (LoggerParameter.USE_PLATFORM_CACHE == false || getTransactionCache().contains(key)) {
```
jamessimone, line 83: it _feels_ like there's still a chance for further consolidation of the `get` checks between the `OrganizationCache` and `SessionCache` cacheables, though the Session one does an additional check - not sure how worth it doing something like that would be though, this is probably fine as is

nebula-logger/extra-tests/tests/LogManagementDataSelector_Tests_Flow.cls
```diff
@@ -0,0 +1,38 @@
...
+// See LICENSE file or go to https://github.com/jongpie/NebulaLogger for full license details.    //
+//------------------------------------------------------------------------------------------------//
+
+@SuppressWarnings('PMD.ApexDoc, PMD.CyclomaticComplexity, PMD.ExcessiveParameterList, PMD.MethodNamingConventions')
+@IsTest(IsParallel=false)
+private class LogManagementDataSelector_Tests_Flow {
+    private static final String TEST_FLOW_NAME = 'MockLogBatchPurgerPlugin';
+
+    @IsTest
+    // static void it_returns_matching_apex_classes_for_specified_class_names() {
```
jamessimone, line 12: oops

nebula-logger/extra-tests/tests/Logger_Tests_Network.cls
```diff
@@ -3,15 +3,11 @@
...
- * @group Extra Tests
- * @description Additional integration tests for orgs with Experience Sites (Communities) enabled
- */
 @SuppressWarnings('PMD.ApexDoc, PMD.CyclomaticComplexity, PMD.ExcessiveParameterList, PMD.MethodNamingConventions, PMD.NcssMethodCount')
 @IsTest(IsParallel=false)
-private class Logger_Tests_ExperienceSite {
-    private static final Boolean IS_EXPERIENCE_CLOUD_ENABLED = Schema.getGlobalDescribe().containsKey('Network');
-    private static final String GUEST_USER_PROFILE_NAME = 'Logger Test Site Profile';
+private class Logger_Tests_Network {
+    private static final Boolean IS_EXPERIENCE_CLOUD_ENABLED = Type.forName('System.Network') != null;
```
jamessimone, line 9: I thought it was `NetworkMember` that provided the crucial piece of info?

scripts/build/enable-debug-mode.apex
```diff
@@ -0,0 +1 @@
+update new User(Id = UserInfo.getUserId(), UserPreferencesUserDebugModePref = false);
```
jamessimone, line 1: aw yeah

jamessimone (approved):
@jongpie just a few comments here and there. Overall this looks great and it's awesome to see all these optimizations ready to go!

nebula-logger/core/main/configuration/classes/LoggerCache.cls
```diff
@@ -0,0 +1,230 @@
...
+    private static final Integer ORGANIZATION_CACHE_TTL_SECONDS = 86400; // 86,400 seconds == 24 hours, max allowd for org cache
+    private static final Cache.SessionPartition SESSION_CACHE_PARTITION = Cache.Session.getPartition(CACHE_PARTITION_NAME);
+    private static final Integer SESSION_CACHE_TTL_SECONDS = 28800; // 28,800 seconds == 8 hours, max allowd for session cache
+
+    private static final OrganizationCache ORGANIZATION_CACHE_INSTANCE = new OrganizationCache();
+    private static final SessionCache SESSION_CACHE_INSTANCE = new SessionCache();
+    private static final TransactionCache TRANSACTION_CACHE_INSTANCE = new TransactionCache();
+
+    @SuppressWarnings('PMD.ApexDoc')
+    private interface Cacheable {
```
jongpie, line 25: @jamessimone I meant to ask you about this earlier. Right now, the `Cacheable` interface is private, and each of the static methods returns a concrete class that implements the interface (and the 3 concrete inner classes are public - `OrganizationCache`, `SessionCache` & `TransactionCache`).

I'm thinking of instead making the `Cacheable` interface public, updating the static methods' returns types to `Cacheable` (instead of the concrete classes), and then make the 3 concrete inner classes `private`. Thoughts on this potential change? I conceptually like the idea because it would completely abstract away the internal implementations - external code would only know about the interface.

nebula-logger/extra-tests/tests/Logger_Tests_Network.cls
jongpie: ....whoops! You're right, thanks for catching this! `NetworkMember` is the useful one to check - I have this check duplicated in a few classes & missed updating this one. I'll have to eventually add a central location for this check so I can cut out the duplication.

jamessimone: cool!

nebula-logger/core/main/configuration/classes/LoggerCache.cls
jamessimone: That would be a great change!

jongpie: Alright, then I'm going for it!

nebula-logger/core/main/configuration/classes/LoggerCache.cls
jongpie: Ooooh, that's a good point, they're very similar, I think I can consolidate it into 1 class. The extra check in the Session cache for `isAvailable()` also exists for Org partitions because they both extend the `Partition` class, and [`Partition.isAvailable()`](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_class_cache_Partition.htm#apex_cache_Partition_isAvailable) seems to always return true for instances of `OrgPartition`, so it doesn't hurt for that extra method check to run in that situation. I'll try to consolidate to 1 class that just has references to `Partition` variables (instead of both `OrgPartition` and `SessionPartition`), I think that'll cut out all of the duplication.

jamessimone: I went to the docs to try to find the shared class - remembered it existed, failed to find it, and left 🤦‍♂️. I was going to suggest an always true check for the org one, but that felt like too much. Glad to know they share a class that can be used to consolidate these checks further!

nebula-logger/core/main/configuration/classes/LoggerCache.cls
jamessimone: makes sense - just wanted to make sure!

jamessimone: I guess the question is - does it make sense for the transaction cache to be able to support nulls if the other two can't? Probably, unless you wanted to consolidate that null check. (Which maybe you can do anyway using the shared `Partition` variable?)

jongpie: Yeah, I think the transaction cache needs to support nulls, there are a few cached queries that return null in some scenarios (like when querying `AuthSession` during async jobs where there isn't an active `AuthSession`). I want to be able to cache the value even when null so that the queries don't constantly re-run - I guess there are other ways I could handle this, but having the transaction cache support null values seemed like the easiest & most consistent approach. 

And I think that duplicate null check is almost eliminated! I'm finishing up some of the consolidation to use the `Partition` class, it looks like that's going to work out really well.

nebula-logger/core/main/configuration/classes/LoggerCache.cls
jongpie: Yeah, I honestly didn't expect them to share an interface or super class, this was a really nice surprise 🥳

jongpie: Ok, I finished making this change (and other changes based on your other comments), and I like this much better! [Here's the commit](https://github.com/jongpie/NebulaLogger/commit/7fc2727d9bd62c75deb723da826e30a260e08d45) for these changes, if you're interested is seeing the final result ("final" 😆 )

nebula-logger/core/main/configuration/classes/LoggerCache.cls
jongpie: Done!

nebula-logger/core/main/configuration/classes/LoggerCache.cls
rmccu: I think you wouldn't set it to null, you would remove the cache key, and then if you do a get, it should return null if the key doesn't exist I think.

jamessimone: to @jongpie 's point, that would counteract the positive effect of transaction caching for some results. There's a big difference between `containsKey(key)` being false, and it being `true` with a null value; in the former, you don't know whether or not you've attempted to cache something, whereas with the latter you know you've cached something and the result was "empty"

rmccu: Yeah, I understand the difference but I wasn't sure if it mattered. If so, you could use a predefined value to represent a `null` state and just check for that when you retrieve the value. By the same token, the cache 'get' method probably shouldn't return null if the key is unset or the value is in the 'null' state. You may want to know that you tried to cache the value already and there is no sense attempting to generate it. This may require you to do something akin to this:
```
private static final String CACHE_IS_NULL = 'CACHE_IS_NULL';

...

Object oCache = LoggerCache.get(key);
if (oCache == null)
{
    // cache needs to be initialized
}
else if (LoggerCache.CASH_IS_NULL.equals(oCache)
{
    // cache was initialized, treat as null value though...
}
else
{
    // cast oCache and use as needed
}
```

jongpie: @rmccu thanks for the sample code! @jamessimone is correct that I don't want to remove the key when the value is `null`, but I do like the idea of a predefined value to represent `null` within platform cache - that seems like a clever way to handle it. I'll do a little prototyping with that approach to see how well it might work. I might still move forward with the current implementation for this PR as-is for now & implement the `CACHE_IS_NULL` later, but let me see what I can come up with in a quick prototype of this change.

jongpie: @rmccu this turned out to be a really nice & easy change, so I went ahead and committed the changes [here](https://github.com/jongpie/NebulaLogger/pull/394/commits/ae3db6fa7fa85bb213b8b988ff210e9578e86066). I was able to keep everything encapsulated within the `LoggerCache` class, so it handles the conversion to/from `null` to the substitute `String` value.

Classes that want to use caching can do something like this to know if a key has been cached/needs to be cached:

```java
public static SomeDataType getCachedValue() {
    String cacheKey = 'someKey';
    // Cache was initialized if contains(key) returns true - even if a null value was cached
    if (LoggerCache.getOrganizationCache().contains(cacheKey)) {
        // At this point, the code knows for sure that the specified key has been previously cached, 
        // so it just needs to cast & return the value. Internally, LoggerCache now handles dealing 
        // with nulls within Platform Cache - if a null value had been previously cached for
        // the specified key, then the Org Partition has the key stored with the substitute String value,
        // and LoggerCache.getOrganizationCache().get(cacheKey) returns null (not the substitute String value)

        // cast cached value and use as needed
        return (SomeDataType) LoggerCache.getOrganizationCache().get(cacheKey);
    } else {
        // cache needs to be initialized
        SomeDataType value; // TODO populate value
        LoggerCache.getOrganizationCache().put(cacheKey, value);
        return value;
    }
}
```

rmccu: Awesome! Glad it worked out smoothly. We did stuff like this at another company where we used the lower boundary values of an int (like negative 2 billion and change) to represent INT_NULL and INT_NOT_SET. It's not perfect, but it works.

I did have a side-question. Do you need to know the difference between null being initialized and not-initialized? For instance, if it is null because it is not initialized, that mean you need to go fetch the value to store and "initialize" the cache entry. Vs, when you know it is initialized and is null, you don't need to fetch.

I have had this problem with platform cache before. I think I punted and just took null to mean I need to initialize. This may incur additional cost if the value is null, but it was an edge case for me at the time.

rmccu: Doh! Just read the last part of 2nd message. We are good.

jongpie: Awesome, thanks @rmccu! I need to wrap up a little more testing & fix an unrelated failing test, but I'm hoping to finally get this PR merged & released today or tomorrow.

nebula-logger/core/main/configuration/classes/LoggerCache.cls
```diff
@@ -182,8 +188,11 @@ public without sharing class LoggerCache {
         public void put(String key, Object value) {
             this.transactionCache.put(key, value);
 
-            // Platform cache does not support storing null values
```
jamessimone: maybe it's just me, but I think I prefer this approach over the committed change here. `null` as a case to be handled is a fairly common thing, whereas by using a sentinel value and hot-swapping it in the `get`, we're adding overhead to _every_ cached value we attempt to access.

jongpie: Just to follow up after we chatted - I'm going to stick with the current approach of using the sentinel value since there's some goofiness with platform cache, the `LoggerCache` internally still prefers using the transaction cache (even when platform cache is enabled & available). Thanks again for talking through this!

PR #395, opened 2022-10-18
OriginLocation__c as (indexed) ExternalId for reporting

JeroenSfdc:
Allows better reporting, summarizing over OriginLocation__c (efficiently)

### DIFF
--- a/nebula-logger/core/main/log-management/objects/LogEntry__c/fields/OriginLocation__c.field-meta.xml
+++ b/nebula-logger/core/main/log-management/objects/LogEntry__c/fields/OriginLocation__c.field-meta.xml
@@ -3,7 +3,7 @@
     <fullName>OriginLocation__c</fullName>
     <businessStatus>Active</businessStatus>
     <complianceGroup>None</complianceGroup>
-    <externalId>false</externalId>
+    <externalId>true</externalId>
     <label>Origin Location</label>
     <length>255</length>
     <required>false</required>


jongpie:
@JeroenSfdc thanks for looking at this, but unfortunately, I can't support this change - the `externalId` attribute for fields cannot be changed in managed packages ([source](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/packaging_packageable_components.htm#mdc_custom_field) & shown in screenshot below). This is a frustrating limitation, but since I'm supporting both a managed package & unlocked package for Nebula Logger, I want to keep the metadata as consistent as possible.

image

I'm unfortunately going to have to close this pull request for now, but I'll log an issue to see if there are any ways to support indexing the data in the future. One possible solution is to deprecate the `OriginLocation__c` field and replace it with a new text field that's flagged as an external ID - this could cause a headache in a lot of orgs that already use Nebula Logger (they would have to update list views, reports, etc. to use the new field), but I'm still going to consider it as a possible solution.

For your specific use case, can you share any details on issues you're running into? I'm guessing that reports & list views in your org might be slow when filtering on `OriginLocation__c`, but could you share some specifics on the kinds of issues you're facing? Are things slow to load? Or do they timeout entirely?

JeroenSfdc:
> @JeroenSfdc thanks for looking at this, but unfortunately, I can't support this change - the `externalId` attribute for fields cannot be changed in managed packages ([source](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/packaging_packageable_components.htm#mdc_custom_field) & shown in screenshot below). This is a frustrating limitation, but since I'm supporting both a managed package & unlocked package for Nebula Logger, I want to keep the metadata as consistent as possible.
> 
> image
> 
> I'm unfortunately going to have to close this pull request for now, but I'll log an issue to see if there are any ways to support indexing the data in the future. One possible solution is to deprecate the `OriginLocation__c` field and replace it with a new text field that's flagged as an external ID - this could cause a headache in a lot of orgs that already use Nebula Logger (they would have to update list views, reports, etc. to use the new field), but I'm still going to consider it as a possible solution.
> 
> For your specific use case, can you share any details on issues you're running into? I'm guessing that reports & list views in your org might be slow when filtering on `OriginLocation__c`, but could you share some specifics on the kinds of issues you're facing? Are things slow to load? Or do they timeout entirely?

Thanks for the super fast turn around with the explanation @jongpie!
The main rationale here is that summarized reports, grouping errors from the same context/origin is something I'm very much used to in current non-Nebula implementations. It makes it easier to identify which issues occur with a given frequency, from a given context/origin. And making reports/dashboards work well on high(er) volumes of data, the indexed field can make/break the report/dashboard. 

Alternatively having an `OriginLocationIndexed__c` which is set with the same value could be an option as well. That would be least-intrusive. But I can imagine it's not a path you'd like to walk with the package.

jongpie:
@JeroenSfdc thanks for explaining the rationale! I think that all makes sense, and in hindsight, I wish that I had originally flagged this field as an external ID - reporting/filtering on this field for large data volumes could definitely be problematic for a lot of orgs.

I'm still not 100% sure how I'm going to try to handle this, but I've created a new issue #400 to track this - hopefully I can implement a solution in the next few months.

Issue #280, opened 2022-02-23
Fail to create logs on wire properties


jefersonchaves:
### Package Edition of Nebula Logger

Unlocked Package

### Package Version of Nebula Logger

4.6.16

### New Bug Summary

First of all thanks for the logger.

In an attempt to log errors on a wire function then it errors out with the following error:
`Cannot read properties of undefined (reading 'supportedLoggingLevels')`

The errors seem to be caused due to the fact the `wire` adapters are not predictable ordered.

I can try to update the component to use imperative retrieval of custom settings instead. This could potentially allow the usage of the logger without requiring to introduce it on markup.

jongpie:
Hi @jefersonchaves - thanks for reporting the issue! I can try to look into this sometime over the next few weeks. Any chance you could provide some sample code that demonstrates the issue? Just want to be sure that I'm taking the right steps to reproduce the problem.

jongpie:
Perfect, thanks so much @jefersonchaves! Once I have a chance to do some testing, I'll let you know if I have any follow-up questions

jefersonchaves:
Hello @jongpie : I recently got some time to work on this. I can submit a PR tomorrow or early next week with my proposal for the LWC logger that will allow usage on wire methods and any other. The solution I plan to show you will enable use without adding the markup.

PR #401, opened 2022-10-28
Logger LWC Enhancements

jefersonchaves:
It solves #280 allowing greater flexibility and not requiring adding the logger on the markup.

I end up creating a new component because this would be a breaking change.

This allowed me to promisify the properties and tidy some things up.

I hope this is a good addition and it would be great to use this in more projects.

Feel free to reach out and/or make changes.

Note: I have also solved a little inconsistency on the LWC Logger demo page about logging with the finest logging level.


jongpie:
@jefersonchaves I did a quick review, and I really seems promising (ha!....? 😆). I need to wrap up 2 other PRs before I can do a thorough review & testing of your changes, but I'm hoping I can take a closer look within the next week. Offhand, my main concern is that it is a new component - I'd really like to only have 1 component if possible, I have a few ideas on how it could maaaaybe be done as part of the existing component without being a breaking change (I'll need to do some prototyping to confirm though). 

I'll let you know as soon as I've had a chance to take a closer look, thanks so much for working on this!

jefersonchaves:
Hey, of course. That is already a great start.

I have made a new component because it would be a breaking change to adopt this in replacement to the wire.

This is proposal and I'm really open for feedback (already feeling flattered that you say it looks promising).

jongpie:
Hey @jefersonchaves! Thanks for your patience, I've finally had a chance to take a closer look at this PR. I really like the changes overall, but I'm running into 1 issue with saving the log when trying to test this. Here are the steps I took:

1. I checked out your branch and updated the `loggerLWCDemo.js` file to have all of the LWC lifecycle hooks - each event adds some log entries & calls `this.logger.saveLog()`

  ```javascript
    constructor() {
        super();
        console.log('>>> start of constructor()');
        this.logger.info('>>> running constructor(), using lwcLogger');
        this.logger.info('>>> adding an extra log entry');
        this.logger.saveLog();
        console.log('>>> done with constructor()');
    }
    
    connectedCallback() {
        console.log('>>> start of connectedCallback()');
        this.logger.saveLog();
        this.logger.info('>>> running connectedCallback(), using lwcLogger');
        this.logger.info('>>> adding an extra log entry');
        this.logger.saveLog();
        console.log('>>> done with connectedCallback()');
    }
    
    disconnectedCallback() {
        console.log('>>> start of disconnectedCallback()');
        this.logger.info('>>> running disconnectedCallback(), using lwcLogger');
        this.logger.info('>>> adding an extra log entry');
        this.logger.saveLog();
        console.log('>>> done with disconnectedCallback()');
    }
    
    renderedCallback() {
        console.log('>>> start of renderedCallback()');
        this.logger.info('>>> running renderedCallback(), using lwcLogger');
        this.logger.info('>>> adding an extra log entry');
        this.logger.saveLog();
        console.log('>>> done with renderedCallback()');
    }
   ```

2. When I open the `loggerLWCDemo` component, all of the log entries appear in the browser's console, but the calls for `saveLog()` execute first, so the new log entries aren't automatically saved (if you then click the "Save Log" button on the LWC, the log entries do successfully save, but ideally the call to `saveLog()` within the lifecycle hooks should also work). 

I looked at your changes in the `saveLog()` function and [it looks like you're already calling `Promise.all()` before calling `saveComponentLogEntries()`](https://github.com/jefersonchaves/NebulaLogger/blob/eb52af0ac4ef56c8ccf2190a723e6ef5ee8a8b97/nebula-logger/core/main/logger-engine/lwc/lwcLogger/lwcLogger.js#L146-L147), which I thought would have handled this, so I'm not 100% sure what the issue is. Let me know if you have any thoughts on what would be causing the issue with saving.

In the meantime, I'll continue reviewing & testing your changes - I'll let you know if I have any other questions!

jefersonchaves:
Hey @jongpie,

Thanks for this review. Fair enough.

Now because `saveLog` is a promise, to guarantee it executes before the next `console.log` we will have to call `await`.

I have not added because the initial sample was saving only when the button was pressed but I can have a look, is just uglier but a good showcase.

jefersonchaves:
To be honest, let me settle that in because that is a fair question and my javascript skills are not the best.

jongpie:
Hi @jefersonchaves, apologies, I had overlooked your last response! But if you'd like, I can try to work on making some additional changes in your branch - I'll have some free time this month, so I'm happy to help out on this PR if that would be useful. Let me know what you think!

jefersonchaves:
hey @jongpie - Thanks, yes, that is indeed very helpful and kind of you. I'm unable to prioritize this right now so your help would be fantastic.

jongpie:
@jefersonchaves that sounds great, I'm happy to help! I'll make some changes in your branch, and I'll let you know when everything is ready - I'm hoping that we can have this merged by the end of December/early January.

jefersonchaves:
Yep, sounds like a plan.

jongpie:
@jefersonchaves thanks so much for your patience over the last several months! I had hoped to work on this in January but had some unexpected home issues come up over the last few months that caused further delays. 

I've finally revisited your changes, and overall this still looks great! I've committed & pushed a couple of changes to your branch:
1. Merged in the latest changes from `main` (v4.10.0) into your branch
2. Combined your `lwcLogger` component with the original `logger` component so that there is still only 1 LWC used for logging. To support this, I made a few small changes:
   - Your version of `lwcLogger/lwcLogger.js` is now stored in `logger/loggerService.js` (with a few minor changes)
   - The original `logger.js` file still extends `LightningElement` and uses `@api` functions so that code in existing orgs (that use `logger` by adding it to their markdown) will continue to work. This avoids a breaking change for existing orgs/code. 
   - Each `@api` function in `logger.js` now delegates functionality to the equivalent function in `loggerService.js`
   - Using `logger` can be done by importing the function `getLogger()` - you had originally called this `createLogger()`, I ended up renaming it, but it's still the same idea as your original implementation
3. Scope creep: I made some additional enhancements in `logEntryBuilder.js` that improve the formatting of messages shown when Nebula Logger calls `console.log()`
4. I've added all lifecycle events (with logging) to the component `loggerLWCDemo`, just to provide a complete example of logging.

If you have a chance, let me know your thoughts on these changes! I'm planning to do some more testing over the next few days, and will hopefully (finally!) merge this PR next week. Thanks again for all of your help!

jefersonchaves:
Hey @jongpie,

Thanks a lot for the review and changes but nonetheless to come back, Nebula Logger is awesome.

All the changes make sense, my rationale behind the dichotomy between `get` and `create` logger is because I felt that this was more complicated than simply returning a value - but look, this is more about semantics.

I loved that you have found a way to keep backward compatible - this one was very tricky and almost kept me awake at nigh.

I will do a bit of testing and let you know a more in-depth review - hopefully on Monday.

jongpie:
That sounds great @jefersonchaves - looking forward to hearing your feedback after you've had a chance to do a more in-depth review. Thank you!

nebula-logger/core/main/logger-engine/lwc/logger/logEntryBuilder.js
```diff
@@ -107,28 +95,40 @@ const LogEntryBuilder = class {
...
+                    consoleLoggingFunction = console.debug;
+                    break;
+            }
+
+            consoleLoggingFunction(consoleMessagePrefix, consoleFormatting, this.message);
+            console.groupCollapsed(consoleMessagePrefix, consoleFormatting, 'Details for: ' + this.message);
+            consoleLoggingFunction(JSON.parse(JSON.stringify(this)));
+            console.trace();
+            console.groupEnd();
+        });
```
jefersonchaves, line 128: This is pretty neat, it is scope creep, but it makes sense.

nebula-logger/core/main/logger-engine/lwc/logger/logEntryBuilder.js
```diff
@@ -107,28 +95,40 @@ const LogEntryBuilder = class {
...
+                    consoleLoggingFunction = console.info;
+                    break;
+                default:
+                    consoleLoggingFunction = console.debug;
+                    break;
+            }
+
+            consoleLoggingFunction(consoleMessagePrefix, consoleFormatting, this.message);
+            console.groupCollapsed(consoleMessagePrefix, consoleFormatting, 'Details for: ' + this.message);
+            consoleLoggingFunction(JSON.parse(JSON.stringify(this)));
```
jefersonchaves, line 125: The cloning `JSON.parse(JSON.stringify`, is never perfect (it does not copy functions, and fails on cyclic objects), so it might be worth a comment to clarify the intent (especially because it is about the output into the browser.

nebula-logger/core/main/logger-engine/lwc/logger/logger.js
```diff
@@ -122,58 +107,37 @@ export default class Logger extends LightningElement {
...
-
-        return logEntryBuilder;
+        this.#loggerService.saveLog(saveMethodName);
     }
 }
+
+/**
+ * @return {LoggerService} a LoggerService instance
+ */
+const getLogger = function () {
```
jefersonchaves, line 136: I suggest createLogger to keep consistent with the `createLoggerService` because it clarifies the intent with some work involved other than just fetching a reference and/or pointer.

```suggestion
const createLogger = function () {
```

jefersonchaves:
Hello @jongpie,

No worries, I get you, and thanks for bearing with me (crazy times at home and other personal stuff for me too).

I have added a few comments that can help with readability, but overall it looks great (thanks a lot for the hard work, it has been my pleasure to enhance it).

⚠️  **Important**: we are breaking code that previously relied on `getUserSettings` to change configuration values. This is by design and was my intent, but it is good to highlight it on the release notes to avoid frustration.

jefersonchaves (commented):
Hello @jongpie,

No worries, I get you, and thanks for bearing with me (crazy times at home and other personal stuff for me too).

I have added a few comments that can help with readability, but overall it looks great (thanks a lot for the hard work, it has been my pleasure to enhance it).

⚠️  **Important**: we are breaking code that previously relied on `getUserSettings` to change configuration values. This is by design and was my intent, but it is good to highlight it on the release notes to avoid frustration.

nebula-logger/core/main/logger-engine/lwc/logger/logEntryBuilder.js
jongpie: Great suggestion, I'll add a comment to document this - long-term, there may be better way(s) to print out the data, but a comment for now should help.

nebula-logger/core/main/logger-engine/lwc/logger/logEntryBuilder.js
jongpie: Yeah, I didn't want to scope-creep too much, but this was something I've wanted to do for a while to help make the console output look nicer & easier to read

nebula-logger/core/main/logger-engine/lwc/logger/logger.js
jongpie: My concern with using `create` is that it's too specific - other components should know or care how the logger instance is being returned. I think that `get` is a more generic verb that helps (purposefully) abstract away the implementation, and would still be accurate if there are ever future enhancements where it _would_ make sense to return a reference/pointer to a singleton instance of logger.  But `createLogger` wouldn't be accurate anymore if (in future releases), references/caching/singleton are ever used.

jongpie:
@jefersonchaves thanks for the feedback! Regarding the breaking change for `getUserSettings`, what was your thoughts behind it? In the Apex method `Logger.getUserSettings()` and previous versions of `logger.js`, I've made the objects modifiable so that settings can be modified in-memory by plugins/other automations. I'm hesitant to switch it to read-only (as some people might be relying on it being modifiable), but would love to know your thoughts!

nebula-logger/core/main/logger-engine/lwc/logger/logger.js
jefersonchaves: Look, fair enough, sometimes I get into semantics because I feel they matter but let's not get hung up on that and move forward.

jefersonchaves:
> @jefersonchaves thanks for the feedback! Regarding the breaking change for `getUserSettings`, what was your thoughts behind it? In the Apex method `Logger.getUserSettings()` and previous versions of `logger.js`, I've made the objects modifiable so that settings can be modified in-memory by plugins/other automations. I'm hesitant to switch it to read-only (as some people might be relying on it being modifiable), but would love to know your thoughts!

@jongpie,
I'm thinking about the Mayers Open-closed principle (from SOLID). The risk is the ability of the logger library to evolve, given unintended dependency (they are plenty of examples out there of the software being bitten by that). If there are things we want to allow consumers to change, we should keep them explicit instead of relying on the pure data structure.

If consumers were changing configurations on the fly, they probably needed configs (to be honest, the test was one example). And the apex `getUserSettings` is not fully read-write either; the consumer still needs to update like this code (I'm yet to see a good plugin code that is updating internal data structures like these - not even current Nebular Logger Plugins do that as far as I can see) https://github.com/jongpie/NebulaLogger/blob/d090f6a8d7a6814ce2ee4bb01c2cab8720abf5f9/nebula-logger/plugins/big-object-archiving/plugin/classes/LogEntryArchivePlugin_Tests.cls#L146.

One course of action might be to deprecate the usage with a warning while keeping it read-write - quite a safe bet. Then on the major, make it read-only on both (for the apex, we can send back a cloned reference instead of the actual config pointer).

jongpie:
@jefersonchaves I see your point - let's move forward with the logger LWC returning a read-only object from `getUserSettings()`, and I'll evaluate making a similar change in Apex separately. I think the Apex method is being used by some orgs to aid in testing, so I want to do some more analysis on the impact of changing it there. For the LWC, although it will be a breaking change, I don't think it will be as impactful of a change - I'll add a note to the release notes about this change, and then I'll see if others report it as an issue for their usage.

With that in mind, I think this PR is more or less ready - I just need to finish some cleanup & add some comments, and then I'll manually create the package version (the pipeline unfortunately doesn't work on forked repos, so I'll just manually create the package version). I hope to have this finished & merged this week.

jefersonchaves:
Thanks @jongpie.
That makes sense, and hey, great collab, I loved it.
If there is more that I can contribute, please let me know. It has been a pleasure and I reap a lot of value from the tool and mostly your work!

jongpie:
@jefersonchaves I will definitely let you know if anything else comes up with this PR. And thanks again for all of your help & patience in getting this finished, it's been great collaborating on this & I'm really excited about these changes, I think they'll really help improve logging for lightning components.

nebula-logger/core/main/logger-engine/lwc/logger/logger.js
jongpie: @jefersonchaves I was out on vacation & didn't have a chance to merge this PR yet - after reviewing it again,I think you're right, using `createLogger()` does help clarify the intent, so I'm reverting from `getLogger()` back to `createLogger()`.

jongpie:
@jefersonchaves it's finally happening - I'm merging this PR today! I made 3 more commits that will be included:
- I've reverted back to using the function name `createLogger()` (instead of `getLogger()`)
- I made the function `loadSettingsFromServer()` pseudo-private and instead updated `getUserSettings()` to accept a parameters object, which includes the property `forceReload`. This should make it easy to add & removed named parameters to `getUserSettings()` in the future (if additional parameters are later needed)
- I added some additional Jest tests to ensure that the now-deprecated markup approach still works correctly

I'm manually running the pipeline locally (I don't currently have the GitHub pipeline setup to work correctly on forked repos), and then I'll merge this later today.

Thanks again so much for all of your help!!

jamessimone:
@jefersonchaves I'm very happy to let you know we've started using the `createLogger` factory function within one of our new LWCs, as well. From myself, @jongpie, and the rest of our team here at Salesforce, thanks again for contributing this functionality!

jefersonchaves:
Thanks @jamessimone - I'm happier to see more people benefiting. And I'm sure all our paths will cross, one PR at a time for now.

PR #407, opened 2022-11-21
Track when an admin user logs in as another user

### LINKED ISSUE #396: Track when another user is impersonating the current user
Salesforce provides ways for users to login as (impersonate) another user, which can be done in 1 of 2 ways:
  1. System admins use the feature [Administrators can log in as any user](https://help.salesforce.com/s/articleView?id=000386710&type=1), which allows any admin to login as another other active user in the org.
  2. Delegated admins use the feature [Enable Group for Login Access](https://help.salesforce.com/s/articleView?id=sf.delegating_user_administration.htm&type=5) to login as another active user within the role hierarchy that they administer.

In these situations, Nebula Logger will store the impersonated user in `Log__c.LoggedBy__c`, with no indication that the user was being impersonated by another user (the impersonator). @surajp had a great suggestion [on SFXD](https://discord.com/channels/246568944213819393/348054624256786434/1034472127568560209) to have this information tracked to better help admins & developers understand how a `Log__c` was generated.

There needs to be a way to identify situations where the user is being impersonated, and have Nebula store the impersonator in a new `User` lookup field `Log__c.ImpersonatedBy__c` (or something similar). Based on some quick testing, it looks like this can be determined when querying the current user's `AuthSession`
  - The `SessionType` field is set to 'SubstituteUser`
  - The `UsersId` field is still set to the current user (the user being impersonated)
  - The `LoginHistory.UserId` field is set to the impersonator

image

jongpie:
Closes #396 by adding new fields & logic to determine if the current user (stored in `Log__c.LoggedBy__c`) is being impersonated by another user

- Added config & code to track impersonators using the Login As feature - this information is now tracked in the new field `LogEntryEvent__e.ImpersonatedById__c` and `Log__c.ImpersonatedBy__c`
  
  image
  
- Added new class `LoggerSObjectProxy` with inner classes that provide representations of problematic SObject Types `AuthSession`, `LoginHistory` and `Network`. This simplifies some of the downstream code, and makes things easier to test.
- Fixed an unreported `SObjectException` that would occur when logging a record ID for a template object, such as `CaseComment`, `CaseHistory`, `AccountHistory`, etc., with the error `System.SObjectException: Cannot locate Apex Type for ID`


jongpie:
This one is tricky (impossible?) to test via integration tests since you can't use the "Login As" functionality via Apex, so @jamessimone, @nojanv, and Jovan helped me manually test this in a sandbox, and everything appears to be working! I logged in as myself, and they used the "Login As" feature to also login as my user - we each ran a script to log some data, and each person's log was correctly attributed to the right person.

PR #411, opened 2022-11-22
Added a custom index for LogEntry__c.OriginLocation__c

### LINKED ISSUE #400: Provide a solution for indexing the field LogEntry__c.OriginLocation__c
thanks for looking at this, but unfortunately, I can't support this change - the `externalId` attribute for fields cannot be changed in managed packages ([source](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/packaging_packageable_components.htm#mdc_custom_field) & shown in screenshot below). This is a frustrating limitation, but since I'm supporting both a managed package & unlocked package for Nebula Logger, I want to keep the metadata as consistent as possible.

image

I'm unfortunately going to have to close this pull request for now, but I'll log an issue to see if there are any ways to support indexing the data in the future. One possible solution is to deprecate the `OriginLocation__c` field and replace it with a new text field that's flagged as an external ID - this could cause a headache in a lot of orgs that already use Nebula Logger (they would have to update list views, reports, etc. to use the new field), but I'm still going to consider it as a possible solution.

For your specific use case, can you share any details on issues you're running into? I'm guessing that reports & list views in your org might be slow when filtering on `OriginLocation__c`, but could you share some specifics on the kinds of issues you're facing? Are things slow to load? Or do they timeout entirely?

_Originally posted by @jongpie in https://github.com/jongpie/NebulaLogger/issues/395#issuecomment-1286049573_

jongpie:
Fixes #400 by adding a custom index for the field `LogEntry__c.OriginLocation__c`, using the new `CustomIndex` metadatatype. This should help to improve queries & reports that filter on this field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,11 @@ The most robust logger for Salesforce. Works with Apex, Lightning Components, Fl
 
 ## Unlocked Package - v4.10.5
 
-[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000023SC7QAM)
-[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000023SC7QAM)
+[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000023SCHQA2)
+[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000023SCHQA2)
 [![View Documentation](./images/btn-view-documentation.png)](https://jongpie.github.io/NebulaLogger/)
 
-`sfdx package install --wait 20 --security-type AdminsOnly --package 04t5Y0000023SC7QAM`
+`sfdx package install --wait 20 --security-type AdminsOnly --package 04t5Y0000023SCHQA2`
 
 ## Managed Package - v4.10.0
 
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 The most robust logger for Salesforce. Works with Apex, Lightning Components, Flow, Process Builder & Integrations. Designed for Salesforce admins, developers & architects.
 
-## Unlocked Package - v4.10.4
+## Unlocked Package - v4.10.5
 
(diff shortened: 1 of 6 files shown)


jongpie:
I've been doing some testing of the new `CustomIndex` metadata type - overall, it seems to work well, but I'm currently running into issues with trying to delete a custom index.

I've tried deleting with `sfdx` using both the new `source` and older `mdapi` commands, but neither seems to work - they both throw an exception `insufficient access rights on cross-reference id`
  - `sfdx force:mdapi:deploy -d ./customindexDelete -w -1` (assuming you have a folder called 'customindexDelete' with all of the appropriate XML files) 
  
    image

  - `sfdx force:source:delete --metadata CustomIndex:LogEntry__c.OriginLocation__c`
    
    image

I'll continue testing this - it makes me a little nervous to merge this PR before having a rollback plan in place (just in case there's some negative impact in some orgs), so I want to figure out deleting the index before I release this change.

JeroenSfdc:
Agree @jongpie since not documented as such, that fact that deploying the meta-data 'works' (did you run an explain plan to see if truly the field was indexed, not only 'checked' as indexed) is tricky. Risk-reward is not such that this needs to be pushed.

jongpie:
@JeroenSfdc yeah, I ran a few tests before & after deploying the custom index, and it does appear to be actually indexing the field - but given that I haven't found a way yet to delete the custom index, I'm going to hold off on merging this PR for now. 

I'll try reaching out internally at Salesforce to see if anyone can provide me with some more info on the new metadata type - if/when I can figure out how to delete a custom index, then I think I'll move forward with merging this PR.

jongpie:
@JeroenSfdc this is now available in [the `v4.10.5` release](https://github.com/jongpie/NebulaLogger/releases/tag/v4.10.5). I've talked with a few other people that are using the new `CustomIndex` metadata type, and so far, it seems to be going well for everyone, so it made sense to finally merge this PR.

JeroenSfdc:
Nice @jongpie! Since we're on the manage package, will need to wait a bit. Thanks a lot!

jongpie:
@JeroenSfdc, some unfortunate news, I'm not going to include the custom index in the next managed package release (`v4.11.0`)  

I've recently seen a few errors about the custom index in a couple of unrelated orgs that deploy Nebula Logger's metadata (instead of using the unlocked or managed packages). For some reason, the inclusion of the custom index causes a deployment error. It doesn't seem to be a widespread issue (I only know of 2-3 orgs impacted)), and I can't reproduce the issue myself in any of my own orgs, so it could be something specific to those orgs. But as a precaution, I'm going to exclude it from the managed package this release to avoid any upgrade issues. 

In the meantime, I'll continue to investigate these errors further - hopefully, I'll be able to include the index in the next managed package release, [`v4.12.0` (Winter '24)](https://github.com/jongpie/NebulaLogger/milestone/12).

PR #452, opened 2023-02-13
Added support for logging Map<Id, SObject> in core, added support for configurable fields in Slack plugin

### LINKED ISSUE #447: Null pointer Exception

Unlocked Package

v4.7.8

### New Bug Summary

Getting a Null pointer Exception inside  ruleCriteriaMet(LogEntryEventHandler Class) when there is no value for the matching field(Log [Entry). We are using a Log Entry Tag rule based on the Origin__c field and when rule Criteria is excecuted it is having null value and comparison throwing a null pointer Exception.

### LINKED ISSUE #453: Default Logger Settings values change the results of unit tests in sandboxes vs. production
The behavior of `Logger.getUserSettings()` changes between sandbox and production environments, and can cause tests to pass in one and fail in the other unexpectedly if the set log levels are different.

Example:
```java
@IsTest
static void noLogLevelSetByTestOrOrg_PassesSandboxButNotProd() {

	Logger.info('Hello World');
	
	// query for it - works in sandbox but not prod
	List<LogEntry__c> entries = ...;
	Assert.areEqual(1, entries.size());
}
```

In the above example, no log level is set by the test or using org defaults, so `Logger.getUserSettings()` will default to picking up `FINEST` as the log level in sandboxes and scratch orgs - all appears fine because the defaulting code has a specific exception made for sandboxes. 

However if Production has an org default set higher than this, the test will fail.

```java
@IsTest
static void logLevelSetByTest_PassesBoth() {
	Logger.getUserSettings().LoggingLevel__c = System.LoggingLevel.INFO.name();

	Logger.info('Hello World');
	
	// query for it - works in all envs
	List<LogEntry__c> entries = ...;
	Assert.areEqual(1, entries.size());
}
```

...the fix is straightforward. The logging level specified inside the test keeps the behavior consistent between environments. I think the problem here is that it changes behavior in a way we didn't expect. I'd almost rather have it fail the same way throughout than fall back to something for sandboxes only? Worth some debate at least.

jongpie:
#### Apex Logging Improvements

- Added new instance method overload `LogEntryEventBuilder.setRecord(Map<Id, SObject> recordIdToRecord)`
  - For now, I've decided to not add overloads for this in `Logger` directly - I might in a future release, but for now, `setRecord()` can be called by chaining the method to one of Logger's methods, like `Logger.info('some message').setRecord(recordIdToRecord)`
- Updated `System.debug()` calls in `Logger` to start with 'Nebula Logger' instead of just 'Logger' for clarity

#### Admin Improvements

- Updated formula field `LogEntry__c.LoggingLevelWithImage__c` to use a different emoji for INFO (✅) vs lower levels (⚙️ for DEBUG, FINE, FINER, and FINEST)
- Added inline help text to all of the limits-related formula fields on `LogEntry__c` to describe the thresholds used to determine which emojis are shown in the formulas (⛔ ⚠️ ✅)
- Unlocked package-only: Added [`AnimationRule` metadata](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_animationrule.htm) for `Log__c.Status__c` to show confetti when the status changes 🥳 🎊 

#### Bugfixes
- Fixed #447 by updating `LogEntryEventHandler` to have an extra null check when processing `LogEntryTagRule__mdt` rules
- Fixed #453 by simplifying `Logger.getUserSettings()` to always use default field values, and making `LoggerSettingsController` responsible for handling prod-specific settings
  - This keeps the behavior of `Logger.getUserSettings()` consistent in all orgs (instead of prod behaving differently from sandboxes/scratch orgs), while still providing the suggested defaults for a prod org via the `loggerSettings` LWC/custom tab
  - Some default field values on `LoggerSettings__c` have also been updated to provide the most detailed settings by default
  - This bugfix also led to the creation of the new public method `Logger.createSettings()` - this doesn't seem needed in the managed package right now, so `public` seems sufficient (instead of using `global`)

### Slack Plugin Package Changes

- Updated `SlackLoggerPlugin` Apex class to leverage a new field set, `Log__c.SlackNotificationFields`, to support customization of the fields shown in Slack notifications
  - This field set can either be directly modified (although changes may be reverted during upgrades), or by creating & specifying another field set in the custom metadata record `LoggerParameter.SlackLogNotificationFieldSet`
  - The Slack remote site setting has also been renamed from `Slack.remoteSite-meta.xml` to `LoggerSlack.remoteSite-meta.xml` to ensure there's consistency in having some variation of the word "Log" in all metadata. Some orgs unintentionally had a previously-existing remote site setting (also named just "Slack") unintentionally overwritten - this name change should help avoid overwriting existing remote site settings for Slack.


jongpie:
@jamessimone if you still have some time to spare, I think this PR is ready for review! In particular, I'd love to know your thoughts on the changes we chatted about yesterday in `Logger` and `LoggerSettingsController` to address #453.

nebula-logger/core/main/logger-engine/classes/Logger.cls
```diff
@@ -3048,7 +3048,7 @@ global with sharing class Logger {
 
     // Inner class for saving log entries via the REST API (using the current user's session ID)
     private class RestApiSaver {
-        private final String baseURL = System.Url.getSalesforceBaseUrl().toExternalForm() + '/services/data/v53.0';
+        private final String baseURL = System.Url.getSalesforceBaseUrl().toExternalForm() + '/services/data/v56.0';
```
jamessimone, line 3051: shouldn't this be 57?

jamessimone (approved):
@jongpie one small question - no worries if you want to bump that API version in a subsequent release though. I think moving responsibility for what the default logging level is in a sandbox to the controller is a great change.

nebula-logger/core/main/logger-engine/classes/Logger.cls
jongpie: I think I'm going to bump it in the next release (when I release a new version of the managed package), I still haven't bumped the rest of the codebase to `v57.0` yet. I'm also tinkering with making the code dynamically add the org's current API version to the string so that I don't have to manually update this line of code, but I didn't want to scope-creep this PR too much.

jongpie:
@jamessimone awesome, thanks so much for reviewing!

nebula-logger/core/main/logger-engine/classes/Logger.cls
jamessimone: Makes sense! And yeah I feel like the good ol' deserialize trick in this one spot would be a good thing.

PR #524, opened 2023-07-05
Increase coverage for LoggerEngineDataSelector to >75%

YodaDaCoda:
Increases coverage for `LoggerEngineDataSelector` by testing `LoggerEngineDataSelector.getNetworkProxies()`. This class previously had only 69% coverage, causing issues when RunSpecifiedTests.

The way this test is written, it should pass and get the necessary coverage even if there are zero `Network`s in the org (is this even possible?)

Very happy to learn how you'd like this updated to see it merged - I see my code style is a little different to other functions in this class but not quite sure exactly how you'd like it.

### DIFF
--- a/nebula-logger/core/tests/logger-engine/classes/LoggerEngineDataSelector_Tests.cls
+++ b/nebula-logger/core/tests/logger-engine/classes/LoggerEngineDataSelector_Tests.cls
@@ -176,6 +176,23 @@ private class LoggerEngineDataSelector_Tests {
         System.Assert.isNull(returnedUser);
     }
 
+    @IsTest
+    static void it_returns_a_list_of_network_proxies() {
+
+        Map<Id, Network> networkMap = new Map<Id, Network>([
+            SELECT Id
+            FROM Network
+        ]);
+
+        Test.startTest();
+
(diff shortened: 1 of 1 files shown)


jongpie:
@YodaDaCoda I'm going to hold off on merging this PR, there is already [a separate test class](https://github.com/jongpie/NebulaLogger/blob/v4.11.0/nebula-logger/extra-tests/tests/LoggerEngineDataSelector_Tests_Network.cls) that's used in the GitHub pipeline for validating the `Network` queries. I've excluded these tests from the core package for a few reasons:

1. The `Network` object doesn't exist in orgs by default - admins have to enable Experience Cloud (Community Cloud) in the org before you can even query the object, so a strongly-typed query (`SELECT Id FROM Network`) would prevent Nebula Logger from being deployable/installable in orgs without Experience Cloud enabled
2. In Apex tests, you can't create `Network` records, so queries will always return 0 records unless you have an actual Experience Cloud site deployed to your org. In Nebula Logger's pipeline, a test Experience Cloud site is deployed to a scratch org to provide integration testing for the query.

tl;dr - testing `Network` queries is a pain in Apex & has to be done dynamically, but Nebula Logger's pipeline runs some extra tests to validate everything's working as expected

PR #525, opened 2023-07-05
Fix typos

jamesgol:
Typo in the README jumped out at me when I first browsed the repo, decided to grep for it and found another instance.


jongpie:
@jamesgol thanks for catching these typos!

Issue #526, opened 2023-07-09
Log Entry Detail not created using SYNCHRONOUS_DML


jongpie:
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/511

<div type='discussions-op-text'>

<sup>Originally posted by **cbcruz74** June 14, 2023</sup>
Hi, we could not get the Log Entry Tag records to be created using SYNCHRONOUS_DML. Log and Log Entry are being created as expected.
The Log Entry Tag records are created with EVENT_BUS only.

Is this working as designed?</div>

JMercie:
Hey @jongpie, I think I found the problem
<img width="822" alt="image" src="https://github.com/jongpie/NebulaLogger/assets/53920455/e99a484b-f835-4373-8025-2f5cdffd6b0c">

When the operation save method is SYNCHRONOUS_DML, the logEntryEvent does not have a UUID, so the else statement never gets executed. Because of this, "this.tagNames" is always empty when the transaction evaluates "upsertLogEntryTags()" method.

<img width="1215" alt="image" src="https://github.com/jongpie/NebulaLogger/assets/53920455/909dbc55-9ae7-4f97-bbd6-adb0fc46fd29">

The solution may be finding another unique identifier for the map, so when we want to reference it with a .get(), we can assign the value unequivocally to the tag record.

PR #539, opened 2023-08-20
Bugfix 526 LogEntryTags__c not being save with SYNCHRONOUS_DML save method

JMercie:
Initial fix to save tags with SYNCHRONOUS_DML as a save method, as discussed in issue #526; I think this could work to solve it. 

Knowing sometimes we won't have the UUid, set the map key to the inserted records Id, and retrieve the values base on it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,6 +65,7 @@ jobs:
               if: ${{ github.event_name == 'pull_request' }}
               shell: bash
               run: |
+                  npx sf version
                   echo "${{ env.DEV_HUB_JWT_SERVER_KEY }}" > ./jwt-server.key
                   npx sf org login jwt --instance-url ${{ env.DEV_HUB_AUTH_URL }} --client-id ${{ env.DEV_HUB_CONSUMER_KEY }} --username ${{ env.DEV_HUB_BOT_USERNAME }} --jwt-key-file ./jwt-server.key --set-default-dev-hub
               env:
@@ -149,6 +150,7 @@ jobs:
             - name: 'Authorize Dev Hub'
               shell: bash
               run: |
+                  npx sf version
                   echo "${{ env.DEV_HUB_JWT_SERVER_KEY }}" > ./jwt-server.key
                   npx sf org login jwt --instance-url ${{ env.DEV_HUB_AUTH_URL }} --client-id ${{ env.DEV_HUB_CONSUMER_KEY }} --username ${{ env.DEV_HUB_BOT_USERNAME }} --jwt-key-file ./jwt-server.key --set-default-dev-hub
               env:
@@ -210,6 +212,7 @@ jobs:
             - name: 'Authorize Dev Hub'
               shell: bash
               run: |
+                  npx sf version
                   echo "${{ env.DEV_HUB_JWT_SERVER_KEY }}" > ./jwt-server.key
                   npx sf org login jwt --instance-url ${{ env.DEV_HUB_AUTH_URL }} --client-id ${{ env.DEV_HUB_CONSUMER_KEY }} --username ${{ env.DEV_HUB_BOT_USERNAME }} --jwt-key-file ./jwt-server.key --set-default-dev-hub
               env:
@@ -283,6 +286,7 @@ jobs:
             - name: 'Authorize Dev Hub'
               shell: bash
               run: |
+                  npx sf version
                   echo "${{ env.DEV_HUB_JWT_SERVER_KEY }}" > ./jwt-server.key
                   npx sf org login jwt --instance-url ${{ env.DEV_HUB_AUTH_URL }} --client-id ${{ env.DEV_HUB_CONSUMER_KEY }} --username ${{ env.DEV_HUB_BOT_USERNAME }} --jwt-key-file ./jwt-server.key --set-default-dev-hub
               env:
@@ -290,6 +294,11 @@ jobs:
                   DEV_HUB_BOT_USERNAME: ${{ secrets.DEV_HUB_BOT_USERNAME }}
                   DEV_HUB_CONSUMER_KEY: ${{ secrets.DEV_HUB_CONSUMER_KEY }}
                   DEV_HUB_JWT_SERVER_KEY: ${{ secrets.DEV_HUB_JWT_SERVER_KEY }}
+
+            # TODO remove this after the plugin @salesforce/packaging has been upgraded
+            # to a new version that only looks at package directories
+            - name: 'Temporarily delete problematic metadata'
+              run: rm -rf ./config/experience-cloud
 
             - name: 'Create Beta Managed Package Version'
               run: npm run package:version:create:managed
@@ -323,6 +332,7 @@ jobs:
             - name: 'Authorize Dev Hub'
               shell: bash
               run: |
+                  npx sf version
                   echo "${{ env.DEV_HUB_JWT_SERVER_KEY }}" > ./jwt-server.key
                   npx sf org login jwt --instance-url ${{ env.DEV_HUB_AUTH_URL }} --client-id ${{ env.DEV_HUB_CONSUMER_KEY }} --username ${{ env.DEV_HUB_BOT_USERNAME }} --jwt-key-file ./jwt-server.key --set-default-dev-hub
               env:
@@ -338,6 +348,11 @@ jobs:
                   npx sf org login sfdx-url --sfdx-url-file ./PKG_DEMO_ORG_SFDX_URL.txt --alias nebula-logger-package-demo
               env:
                   PKG_DEMO_ORG_SFDX_URL: ${{ secrets.NEBULA_PKG_DEMO_SANDBOX_SFDX_URL }}
+
+            # TODO remove this after the plugin @salesforce/packaging has been upgraded
+            # to a new version that only looks at package directories
+            - name: 'Temporarily delete problematic metadata'
+              run: rm -rf ./config/experience-cloud
 
             - name: 'Create & Install Package Version'
               run: npx pwsh ./scripts/build/create-and-install-package-version.ps1 -targetpackagealias '"Nebula Logger - Core"' -targetreadme ./README.md -targetusername nebula-logger-package-demo
@@ -391,6 +406,7 @@ jobs:
             - name: 'Authorize Dev Hub'
               shell: bash
               run: |
+                  npx sf version
                   echo "${{ env.DEV_HUB_JWT_SERVER_KEY }}" > ./jwt-server.key
                   npx sf org login jwt --instance-url ${{ env.DEV_HUB_AUTH_URL }} --client-id ${{ env.DEV_HUB_CONSUMER_KEY }} --username ${{ env.DEV_HUB_BOT_USERNAME }} --jwt-key-file ./jwt-server.key --set-default-dev-hub
               env:
--- a/README.md
+++ b/README.md
@@ -7,13 +7,13 @@ The most robust logger for Salesforce. Works with Apex, Lightning Components, Fl
 
 ## Unlocked Package - v4.11.7
 
-[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y000001HZfGQAW)
-[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y000001HZfGQAW)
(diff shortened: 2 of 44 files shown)


JMercie:
> @JMercie thanks so much for working on this, and apologies for the slow response! I'm hoping to finish reviewing it either today or tomorrow.

Not a problem! Take your time and if it looks good I can start working on the unit tests, I haven't started on it just yet because I'm not sure if this is the best way to approach this.

Sorry if the constant PR feels like spam 😞

jongpie:
Hi @JMercie - please, please, don't feel like your contributions are spam, I'm very grateful for all of your help recently, it's very appreciated! 🙇 

I went through your PR changes, and I think you're definitely correct that the entries without UUID is the problem. One of my concerns with the PR's code is it uses an additional SOQL query (which would need to be maintained as new fields are added). There are also several `if/else` blocks required in the current implementation to check each entry to see if the UUID is null/not null - this isn't complexity that you've introduced, but the existing approach (of relying on `EventUuid__c`) necessitates checking if the UUID is null, and I'd like to reduce some of the code's complexity. I think we can solve the bug by using a composite key for all `LogEntry__c` records, instead of trying to rely on the `EventUuid__c` or `Id` fields - it would simplify some of the code, and all `LogEntry__c` records would be treated consistently.

Right now, there is already an implicit composite key in the data model - the combination of the fields `LogEntry__c.TransactionId__c` and `LogEntry__c.TransactionEntryNumber__c` will always be unique, and both fields will always be populated by Nebula Logger. For example, in `LogEntryEventHandler`, [this block of code](https://github.com/jongpie/NebulaLogger/blob/c014a566f8444298b91e9bb76b76ecba080852fe/nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls#L357-L362)...

```apex
if (logEntry.EventUuid__c == null) {
    logEntriesWithoutUuid.add(logEntry);
} else {
    logEntriesWithUuid.add(logEntry);
    this.logEntryEventUuidToTagNames.put(logEntry.EventUuid__c, logEntryTagNames);
}
```

...could be simplified by using a composite key (with a new map named `logEntryEventCompositeKeyToTagNames`):

```apex
String compositeKey = logEntry.TransactionId__c + logEntry.TransactionEntryNumber__c;
this.logEntryEventCompositeKeyToTagNames.put(compositeKey, logEntryTagNames);
```

This would eliminate several of the `if/else` blocks, and we wouldn't need to requery the `LogEntry__c` records. Any other similar code blocks that currently check if `EventUuid__c` is null would also be updated to use the same composite key approach. Let me know if this approach makes sense/if you have any questions!

JMercie:
> Hi @JMercie - please, please, don't feel like your contributions are spam, I'm very grateful for all of your help recently, it's very appreciated! 🙇 

I went through your PR changes, and I think you're definitely correct that the entries without UUID is the problem. One of my concerns with the PR's code is it uses an additional SOQL query (which would need to be maintained as new fields are added). There are also several `if/else` blocks required in the current implementation to check each entry to see if the UUID is null/not null - this isn't complexity that you've introduced, but the existing approach (of relying on `EventUuid__c`) necessitates checking if the UUID is null, and I'd like to reduce some of the code's complexity. I think we can solve the bug by using a composite key for all `LogEntry__c` records, instead of trying to rely on the `EventUuid__c` or `Id` fields - it would simplify some of the code, and all `LogEntry__c` records would be treated consistently.

Right now, there is already an implicit composite key in the data model - the combination of the fields `LogEntry__c.TransactionId__c` and `LogEntry__c.TransactionEntryNumber__c` will always be unique, and both fields will always be populated by Nebula Logger. For example, in `LogEntryEventHandler`, [this block of code](https://github.com/jongpie/NebulaLogger/blob/c014a566f8444298b91e9bb76b76ecba080852fe/nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls#L357-L362)...

```apex
if (logEntry.EventUuid__c == null) {
    logEntriesWithoutUuid.add(logEntry);
} else {
    logEntriesWithUuid.add(logEntry);
    this.logEntryEventUuidToTagNames.put(logEntry.EventUuid__c, logEntryTagNames);
}
```

...could be simplified by using a composite key (with a new map named `logEntryEventCompositeKeyToTagNames`):

```apex
String compositeKey = logEntry.TransactionId__c + logEntry.TransactionEntryNumber__c;
this.logEntryEventCompositeKeyToTagNames.put(compositeKey, logEntryTagNames);
```

This would eliminate several of the `if/else` blocks, and we wouldn't need to requery the `LogEntry__c` records. Any other similar code blocks that currently check if `EventUuid__c` is null would also be updated to use the same composite key approach. Let me know if this approach makes sense/if you have any questions! 

Hi @jongpie, thanks for all the help and the comments! :D

Using the composite key, code run faster and we also avoid the query and all the clutter just to populate the map in the case the EventUuid is missing. 

I couldn't get to solve how to avoid using this to save the log entry records.

```if (logEntry.EventUuid__c == null) {
    logEntriesWithoutUuid.add(logEntry);
} else {
    logEntriesWithUuid.add(logEntry);
}```

I was thinking on creating a new field to store the composite key (transactionId + transactionentrynumber) and just use the upsert operation using that new field as the externalId field for the method.

But I think this so far solves the problem, tags are getting created with all SaveMethod enums.

Let me know what you think!

**Comment by JMercie** [comment]
@jongpie  Did you find time to look at the solution proposed? Let me know if I need to change anything! 

thank you!

**Comment by jongpie** [comment]
@JMercie I love your idea of creating a new field on `LogEntry__c` to store the composite key - it would simplify the code, and I think it'd also be nice to have the composite key stored for reference.

Let's create a new optional field called `UniqueId__c` on `LogEntry__c`, and populate it with the composite key. Several other objects already have a `UniqueId__c` field that you can mimic, like on [LoggerScenario__c](https://github.com/jongpie/NebulaLogger/blob/v4.11.2/nebula-logger/core/main/log-management/objects/LoggerScenario__c/fields/UniqueId__c.field-meta.xml) or [LoggerTag__c](https://github.com/jongpie/NebulaLogger/blob/v4.11.2/nebula-logger/core/main/log-management/objects/LoggerTag__c/fields/UniqueId__c.field-meta.xml).

Let me know if you have any other questions. The [other changes](https://github.com/jongpie/NebulaLogger/pull/539/commits/2367bb24f70b0c2bbcc32176cca5ba4cb2e47514) that you last made look good to me overall, so I think once you incorporate the new field + unit tests, this should be ready to go!

**Comment by JMercie** [comment]
> @JMercie I love your idea of creating a new field on `LogEntry__c` to store the composite key - it would simplify the code, and I think it'd also be nice to have the composite key stored for reference.
> 
> Let's create a new optional field called `UniqueId__c` on `LogEntry__c`, and populate it with the composite key. Several other objects already have a `UniqueId__c` field that you can mimic, like on [LoggerScenario__c](https://github.com/jongpie/NebulaLogger/blob/v4.11.2/nebula-logger/core/main/log-management/objects/LoggerScenario__c/fields/UniqueId__c.field-meta.xml) or [LoggerTag__c](https://github.com/jongpie/NebulaLogger/blob/v4.11.2/nebula-logger/core/main/log-management/objects/LoggerTag__c/fields/UniqueId__c.field-meta.xml).
> 
> Let me know if you have any other questions. The [other changes](https://github.com/jongpie/NebulaLogger/pull/539/commits/2367bb24f70b0c2bbcc32176cca5ba4cb2e47514) that you last made look good to me overall, so I think once you incorporate the new field + unit tests, this should be ready to go!

Could you please guide me on adequately setting the test to use the Synchronous_DML save method for the LogEntryEvent__e? All other examples in the test class use EventBus from your Mock class. I guess I should rely on getDatabse().insert() somehow.

**Comment by jongpie** [comment]
Hi @JMercie - sure, happy to help! In this situation, I think it's worth having an integration test to ensure that this scenario works. I was reviewing the `LoggerMockDataStore` class, and I'm not sure it would fully validate/replicate the issue of saving with `SYNCHRONOUS_DML`, so using mocks/unit testing may not be robust enough (at least for now).

Here is a small integration test method you can use for reference/modify as you see fit. In the recent versions of Nebula Logger, this test fails due to the line `Logger.saveLog(Logger.SaveMethod.SYNCHRONOUS_DML);` (but passes if you instead use `Logger.saveLog(Logger.SaveMethod.EVENT_BUS);`).

```apex
@IsTest
static void it_should_save_data_when_saving_with_synchronous_dml() {
  System.Assert.areEqual(0, [SELECT COUNT() FROM Log__c], 'Test has started under the wrong conditions');
  System.Assert.areEqual(0, [SELECT COUNT() FROM LogEntry__c], 'Test has started under the wrong conditions');
  System.Assert.areEqual(0, [SELECT COUNT() FROM LogEntryTag__c], 'Test has started under the wrong conditions');
  System.Assert.areEqual(0, [SELECT COUNT() FROM LoggerTag__c], 'Test has started under the wrong conditions');
  String exampleTag = 'Some tag';

  System.Test.startTest();
  LogEntryEventBuilder builder = Logger.info('some INFO message').addTag(exampleTag);
  System.Assert.areEqual(1, Logger.getBufferSize());
  // Logger.saveLog(Logger.SaveMethod.EVENT_BUS); // Using EVENT_BUS, the test passes
  Logger.saveLog(Logger.SaveMethod.SYNCHRONOUS_DML); // Using SYNCHRONOUS_DML, the test fails
  System.Assert.areEqual(0, Logger.getBufferSize());
  System.Test.stopTest();

  System.Assert.areEqual(1, [SELECT COUNT() FROM Log__c], '1 matching Log__c record should have been generated');
  System.Assert.areEqual(1, [SELECT COUNT() FROM LogEntry__c WHERE LoggingLevel__c = :System.LoggingLevel.INFO.name()], '1 matching LogEntry__c record should have been generated');
  System.Assert.areEqual(1, [SELECT COUNT() FROM LogEntryTag__c WHERE LoggerTag__r.Name = :exampleTag], '1 matching LogEntryTag__c record should have been generated');
  System.Assert.areEqual(1, [SELECT COUNT() FROM LoggerTag__c WHERE Name = :exampleTag ], '1 matching LoggerTag__c record should have been generated');
}
```

Let me know what you think, or if you have any other questions!

JMercie:
Hey @jongpie !!

Sorry it took me so long to answer and update the PR. I finally found time to work on this. Thanks a lot for the help with the test method.

I was able to implement the test method successfully; what was making it fail was that the LoggerSObjectHandler.cls class was clearing the list of class handlers retrieved from the LoggerSObjectHandler__mdt metadata type when running on the Test context.

Hopefully, these changes can close this PR. Let me know what you think, and thanks for the patience on this.

jongpie:
Hi @JMercie! Thanks so much for working on those additional changes! I think overall this looks great. There are a few changes that I still want to review a bit more, and then I'll merge this PR sometime this week 🥳

JMercie:
> Hi @JMercie! Thanks so much for working on those additional changes! I think overall this looks great. There are a few changes that I still want to review a bit more, and then I'll merge this PR sometime this week 🥳 

Awesome!! Let me know if everything is fine. Happy to work on any additional changes.

jongpie:
Hi again @JMercie! I finished reviewing the PR, and your overall approach looks fantastic! I ended up making a few small changes while working on creating the new package version
- Merged the `main` branch to your bugfix branch - there were a few other recent releases (after you initially created this branch/PR), so I've updated your branch
- I removed the custom index `LogEntry__c.UniqueId__c.indx-meta.xml` - since the field `LogEntry__c.UniqueId__c` is an external ID, it's automatically indexed
- I updated the field `LogEntry__c.UniqueId__c` from text(50) to text(255) - once I release the field via a new package version, I can't change the length, so using the max value possible (255) gives flexibility in the future if the composite key's structure ever changes
- Reverted the change you made in `LoggerSObjectHandler` - I don't want the code to rely on the metadata in the org because that could cause inconsistent test behavior in orgs that have modified the `LoggerSObjectHandler__mdt` metadata. To make the test work, I added `LoggerTestConfigurator.setupMockSObjectHandlerConfigurations();` to the test method so mock `LoggerSObjectHandler__mdt` records are used.
  - Apologies for forgetting to include this important line of code in the sample test that I had provided you previously!
- Scope creep: Added the new field to 3 of the permission sets, and added it to the `LogEntryRecord` FlexiPage. I also added some conditional visibility rules to some existing fields on the FlexiPage (unrelated to this PR, but they were small issues that were easy to fix/include in this PR)

I think everything is now ready to go! I'm going to work on writing the release notes, and will hopefully merge this later today 🥳

jongpie:
Fixes #526

jongpie (approved):

jongpie:
@JMercie I've merged & release this as [`v4.11.7`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.11.7) 🥳 Thanks again for all of your help!

JMercie:
Awesome!! Thanks a lot; I'm learning much contributing to this repo. 😄

PR #560, opened 2023-09-28
New LoggerAdmin queue as fallback record owner

### LINKED ISSUE #545: Errors occur when new Log__c records are assigned to license types that can't own custom object records
### Reported by @mspiep01 in https://github.com/jongpie/NebulaLogger/discussions/530

By default, new `Log__c` records are created by the Automated Process user and assigned to the `User` that generated the log (by setting `Log__c.OwnerId = Log__c.LoggedBy__c`). This causes an exception when the logging user (`Log__c.LoggedBy__c`) has a license type that can't own object records. [Here are my initial findings](https://github.com/jongpie/NebulaLogger/discussions/530#discussioncomment-6824719) - it seems like the autoproc user, a Chatter Free user, an Identity user, and others profiles will cause (and presumably have been causing) issues.

### Bugfix considerations
When possible, I'd like to still assign the `LoggedBy__c` user as the `OwnerId`. Part of the original intent behind setting `Log__c.OwnerId = Log__c.LoggedBy__c` was to try to [mitigate ownership data skew issues](https://developer.salesforce.com/docs/atlas.en-us.draes.meta/draes/draes_group_membership_data_skew.htm). It also helps in orgs that utilize [the End User permission set](https://github.com/jongpie/NebulaLogger/blob/v4.11.2/nebula-logger/core/main/log-management/permissionsets/LoggerEndUser.permissionset-meta.xml) - `Log__c` has a private sharing model, so end users will only see records that they have been granted access to view, which can be accomplished by assigning them as the record owner.

But when that's not possible, assigning a (new) default queue as the owner seems like the best option. 
- This could still lead to ownership data skew issues (if a large number of `Log__c` records are assigned to the new queue), so some orgs may need to consider manually reassigning (or purging) `Log__c` records if needed. But using a queue would avoid the current issue that some users simply can't own their own records.
- Orgs can customize the queue members directly in their orgs - any changes to queue members made directly in orgs wouldn't be impacted by future upgrades to Nebula Logger
- The queue should include all of Nebula Logger's custom objects. For now, only `Log__c` records will be automatically assigned to the queue, but since there's going to be a queue anyway, might as well make it work for all of the objects in the data model.
- The name of the queue used within Nebula Logger's code should be configurable via `LoggerParameter__mdt`, in case any orgs want to use a different queue

jongpie:
- Fixes #545 (reported by @mspiep01 in discussion #530)

### DIFF
--- a/README.md
+++ b/README.md
@@ -629,7 +629,7 @@ Within Logger Console app, the Log Entry Event Stream tab provides real-time mon
 
 ### View Related Log Entries on a Record Page
 
-Within App Builder, admins can add the 'Related Log Entries' lightning web component (lwc) to any record page. Admins can also control which columns are displayed be creating & selecting a field set on `LogEntry__c` with the desired fields.
+Within App Builder, admins can add the 'Related Log Entries' lightning web component (lwc) to any record page. Admins can also control which columns are displayed by creating & selecting a field set on `LogEntry__c` with the desired fields.
 
 -   The component automatically shows any related log entries, based on `LogEntry__c.RecordId__c == :recordId`
 -   Users can search the list of log entries for a particular record using the component's built-insearch box. The component dynamically searches all related log entries using SOSL.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 The most robust logger for Salesforce. Works with Apex, Lightning Components, Flow, Process Builder & Integrations. Designed for Salesforce admins, developers & architects.
 
-## Unlocked Package - v4.11.6
+## Unlocked Package - v4.11.7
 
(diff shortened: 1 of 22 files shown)


jongpie:
I'm putting this on-hold for the moment - queues can't be included in managed packages, and since Nebula Logger has both an unlocked package and a managed package, the solution needs to work for both. I'll continue researching options to see what's feasible.

jongpie:
After letting this PR sit open for several months, I'm going to close this, and I'm not going to release any changes for #545 at this time:
- The approach used (including a default queue) doesn't work with managed packages, and there isn't a clear alternative approach to use right now that would "just" work out-of-the-box. 
- [It also doesn't seem to be a problem any more](https://github.com/jongpie/NebulaLogger/discussions/530#discussioncomment-9752842). If/when others report this is still an issue, we can revisit figuring out how to handle it.

PR #564, opened 2023-10-04
Store HttpResponse header values

### LINKED ISSUE #561: Log headers from HttpResponse (and possibly HttpRequest) objects
### New Feature Summary

This is an expansion of #240.
Currently the logger logs only header keys for `HttpResponse`, but not their values.
Please consider logging both header keys and values. They could be logged in a simple `Header: value` format (so it matches the [HTTP standard](https://datatracker.ietf.org/doc/html/rfc1945#section-4.2)), and delimited by new lines, just like the keys in the current logger. Retrieving the values requires only calling [`getHeader(key)`](https://developer.salesforce.com/docs/atlas.en-us.apexref.meta/apexref/apex_classes_restful_http_httpresponse.htm#apex_System_HttpResponse_getHeader) method.

Current code (`LogEntryEventBuilder.cls`):
```
this.logEntryEvent.HttpResponseHeaderKeys__c = String.join(response.getHeaderKeys(), NEW_LINE_DELIMITER);
```

New code:
```
private static final String HTTP_HEADER_FORMAT = '{0}: {1}';

(...)

List<String> headers = new List<String>();
for (String key : response.getHeaderKeys()) {
	headers.add(String.format(HTTP_HEADER_FORMAT, new List<String>{key, response.getHeader(key)}));
}
this.logEntryEvent.HttpResponseHeaders__c = String.join(response.getHeaderKeys(), NEW_LINE_DELIMITER);
```
`LogEntryEvent__e.HttpResponseHeaderKeys__c` and `LogEntry__c.HttpResponseHeaderKeys__c` fields would need to be renamed, as after this change not only keys would be logged.

The same could be done with `HttpRequest`, but it'd be tricky, as the current API do not provide `getHeaders()` method.

jongpie:
## New Feature: Store `HttpResponse` Header Values
- Resolved #561 (reported by @TrangOul) by adding new fields `LogEntryEvent__e.HttpResponseHeaders__c` and `LogEntry__c.HttpResponseHeaders__c` that store the full key-value pair for each header set on the logged instance of `System.HttpResponse`. These fields are now automatically set when using the instance method `LogEntryEventBuilder.setHttpResponseDetails(System.HttpResponse)`.
  - Based on [the great suggestion from @TrangOul](https://github.com/jongpie/NebulaLogger/issues/561#issue-1924171609), each header key-value pair is stored in the format `Header: value` to match the ([HTTP standard](https://datatracker.ietf.org/doc/html/rfc1945#section-4.2)). A `\n` line break is used as a delimiter between each header when there are multiple headers.
  - For orgs that do not want to store the header values (either due to security concerns with storing this information, or other concerns), you can disable this functionality org-wide by updating the new `LoggerParameter__mdt` record `StoreHttpResponseHeaderValues` to `false`. By default, this feature is enabled/set to `true`
  - The existing fields `LogEntryEvent__e.HttpResponseHeaderKeys__c` and `LogEntry__c.HttpResponseHeaderKeys__c` will continue to be populated. These fields store only the header keys, not the header values, and some orgs have existing tests & logic that rely on checking the logged keys. These fields are always populated for instances of `System.HttpResponse`, regardless of the configuration of the new `LoggerParameter__mdt` record `StoreHttpResponseHeaderValues`.
- Updated permission sets `LoggerAdmin` and `LoggerLogViewer` to have `read` access to the new field `LogEntry__c.HttpResponseHeaders__c`. 
  - The permission set `LoggerEndUser` intentionally does **_not_** have access to this field (or any other fields related to `HttpResponse` or `HttpRequest`).
- Added new field `LogEntry__c.HttpResponseHeaders__c` to the FlexiPage `LogEntryRecordPage`. This field is shown in addition to the existing field `LogEntry__c.HttpResponseHeaderKeys__c`, shown in the (massive 😅) screenshot below  
  
  image

## Updated Internal Coding Conventions for Apex `if` Statements
- Updated ~300 `if` statements throughout the codebase to simplify checking Booleans from `if(condition == true)` to just `if(condition)`. A few existing Booleans were also updated to have a default value to avoid issues with `null` values.
  - This (hopefully) shouldn't impact any functionality in Nebula Logger - the goal of this change is purely to update the coding style used within the codebase to be a bit more concise, based on some feedback received over the last year or so, including a discussion last year with @rmccu in #383 and [PR comment from @TrangOul](https://github.com/jongpie/NebulaLogger/pull/564#discussion_r1347445950).

### DIFF
--- a/nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
+++ b/nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
@@ -147,7 +147,7 @@ global with sharing class LogEntryEventBuilder {
      */
     global LogEntryEventBuilder setMessage(String message) {
         // To help with debugging unit tests, always run System.debug statement in a test context
-        if ((this.shouldSave() == true || System.Test.isRunningTest() == true) && this.logEntryEvent != null) {
+        if ((this.shouldSave() || System.Test.isRunningTest()) && this.logEntryEvent != null) {
             String cleanedMessage = applyDataMaskRules(message);
             Boolean messageMasked = cleanedMessage != message;
             String truncatedMessage = truncateFieldValue(Schema.LogEntryEvent__e.Message__c, cleanedMessage);
@@ -176,7 +176,7 @@ global with sharing class LogEntryEventBuilder {
         this.logEntryEvent.ExceptionType__c = apexException.getTypeName();
 
         // Stack traces are not returned for managed packages
-        if (isValidStackTrace(apexException.getStackTraceString()) == true) {
+        if (isValidStackTrace(apexException.getStackTraceString())) {
             this.logEntryEvent.ExceptionStackTrace__c = apexException.getStackTraceString();
         }
 
@@ -585,7 +585,7 @@ global with sharing class LogEntryEventBuilder {
         Boolean responseBodyMasked = cleanedResponseBody != response.getBody();
 
         String formattedHeadersString;
-        if (LoggerParameter.STORE_HTTP_RESPONSE_HEADER_VALUES == true) {
+        if (LoggerParameter.STORE_HTTP_RESPONSE_HEADER_VALUES) {
             List<String> headers = new List<String>();
             for (String headerKey : response.getHeaderKeys()) {
                 headers.add(String.format(HTTP_HEADER_FORMAT, new List<String>{ headerKey, response.getHeader(headerKey) }));
@@ -622,7 +622,7 @@ global with sharing class LogEntryEventBuilder {
      * @return The same instance of `LogEntryEventBuilder`, useful for chaining methods
      */
     global LogEntryEventBuilder addTags(List<String> tags) {
-        if (tags == null || tags.isEmpty() == true) {
+        if (tags == null || tags.isEmpty()) {
             return this;
         }
 
@@ -789,7 +789,7 @@ global with sharing class LogEntryEventBuilder {
     }
 
     private void setTagsDetails() {
-        if (this.tags.isEmpty() == true || this.tagDetailsAreSet == true) {
+        if (this.tags.isEmpty() || this.tagDetailsAreSet) {
             return;
         }
 
@@ -848,7 +848,7 @@ global with sharing class LogEntryEventBuilder {
     }
 
     private static void setQueriedNetworkDetails(LogEntryEvent__e logEntryEvent, LoggerSettings__c userSettings) {
-        if (userSettings.IsEnabled__c == false || LoggerParameter.QUERY_NETWORK_DATA_SYNCHRONOUSLY == false || String.isBlank(networkId) == true) {
+        if (userSettings.IsEnabled__c == false || LoggerParameter.QUERY_NETWORK_DATA_SYNCHRONOUSLY == false || String.isBlank(networkId)) {
             return;
         }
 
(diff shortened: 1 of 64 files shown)


nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -583,9 +584,18 @@ global with sharing class LogEntryEventBuilder {
         String cleanedResponseBody = applyDataMaskRules(response.getBody());
         Boolean responseBodyMasked = cleanedResponseBody != response.getBody();
 
+        String formattedHeadersString;
+        if (LoggerParameter.STORE_HTTP_RESPONSE_HEADER_VALUES == true) {
```
TrangOul, line 588: MINOR
Please do not compare with literal Booleans. `if(x)` suffices.

jongpie: Yeah, this has come up a few times, including [this discussion](https://github.com/jongpie/NebulaLogger/discussions/383#discussioncomment-3791831) from ~1 year ago. At the moment, Nebula Logger's codebase currently uses these conventions for all `if` statements:

- `if (someCondition == true)`
- `if (someCondition == false)`

Over the years, I've used Nebula Logger's codebase as a place to experiment with different naming conventions, etc, and a few years ago, I settled on using the above convention of explicitly comparing to literal booleans. But today, I don't use this exact convention in any other Apex code that I work on (besides Nebula Logger). Now, I tend to use:

- `if (someCondition)`
- `if (someCondition == false)`

So, I agree that the extra `== true` piece is unnecessary, Nebula Logger. For `false` conditions, I still like to explicitly comparing to the Boolean `false`, I think it's better for human-readability compared to using an `!` for `false`, like `if (!someCondition)`.

With all that said, I think it's finally time to update Nebula Logger's convention to eliminate the extra `== true` bit (and I'll leave the `== false` lines as-is). I can do it fairly easily with a bulk 'find + replace' in VS Code. There are close to 300 matches in 45 Apex classes of `== true` (see screenshot below), so it's technically a lot of lines of code to change, but I think this is a safe change to make / it shouldn't impact any functionality. I'll look at this more today, and will probably include this change in this PR.

Thanks for the feedback!

image

TrangOul: I can think of one corner case when comparing to a literal Boolean makes sense: by default, all Apex variable are nullable, so actually we can have three values: `false`, `true` and `null`. [The last values causes problems](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/). The following snippet will fail with an NPE:
```
Boolean x;
if (x) {}
```
while
```
Boolean x;
if (x == true) {}
```
will not.

I checked how you set `STORE_HTTP_RESPONSE_HEADER_VALUES` and you make sure that it is never `null`. I haven't checked the other occurrences of `x == true` (or false), though.

I'm glad that you are flexible with coding conventions. They are meant to help up, developers, not to be set in stone.

jongpie: Great point about the `null` values corner case - in general, I try to avoid having `null` Boolean values (by always initializing a default value), but I'll test to confirm that `null` won't cause an issue anywhere in the codebase.

jongpie: @TrangOul you were right! There were a few corner cases with some `null` values that caused some test failures - for newer Apex classes, I've been assigning default values, but there were some older classes in the codebase (used for Flow invocables) that didn't have default values assigned.

I updated a few booleans to assign some default values, and everything seems to be working now 🥳 I've pushed those changes to this branch/PR, and I'm hoping to merge this PR today or tomorrow.

Thanks again!

PR #577, opened 2023-10-23
Properly handle logs created and saved through createLogger in pre-mounted instances

### LINKED ISSUE #518: LWC - Logs not created for some lifecycle events when using new createLogger() pattern
Reported by @shuylern in this discussion: https://github.com/jongpie/NebulaLogger/discussions/515

**UPDATE**

I was able to determine that the saveLog() was actually being executed _prior_ to the logger actually finishing its creation.  

I did this on the wire call which returned a forced Error from the Apex controller.

I added a setTimeout method that paused the saveLog() for 2.5 seconds after the logger creation was initiated.  This resulted in the Log (error) being created without issue.

I am still interested in knowing if there is a delivered solution that I may have overlooked.

```
 @wire(returnSomeString)
    wiredReturnSomeString({ error, data }) {        
        if (data) {
            this.logger.info('>>> wire function return value: ' + data);
        } else if (error) {
            console.log('error in Wire');            
            const entry = this.logger.error('stuck here').setError(error).setRecordId(this.recordId).addTags([this.loggerDeleteByTag,'#LWC']);
            this.error = error;
            this.sleep(2500).then(() => { this.saveLogExample() });
        }
    }

saveLogExample() {
        console.log('running saveLog for btn');
        this.logger.setScenario(this.scenario);
        console.log(this.logger);
        //this.logger.saveLog('QUEUEABLE');
        this.logger.saveLog();
    }

    sleep(ms) {
        return new Promise(resolve => setTimeout(resolve, ms));
    }
```

_Originally posted by @shuylern in https://github.com/jongpie/NebulaLogger/discussions/515#discussioncomment-6308238_

jamessimone:
* Fixes #518 by calling saveLog() as soon as createLogger() has finished loading for pending log messages


jongpie (approved):
@jamessimone thanks again for working on this! Everything looks good to me, I'm going to release this ASAP 🥳

PR #617, opened 2024-01-28
Bugfix: Improved conditional visibility for 'HTTP Response' page section

### LINKED ISSUE #616: Show HTTP Response section on Log Entry Record Page for Apex Continuation
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/586

<sup>Originally posted by **BastienRaoul** December  4, 2023</sup>
Hello @jongpie, 

For one of our older processes, we would like to log the HTTP response of an apex continuation.

The problem we face is that a continuation always returns a null response status.
In the Log Entry Record Page, the HTTP response section is only visible if the response status is not null, therefore we cannot see the section even if the others fields of the HTTP response are not null.

Would it be possible to improve this by creating a dedicated field that is set when at least one of the HTTP response field is not null and use this field for the section visibility ?

Thank you,
Bastien</div>

jongpie:
# Core Package Changes
- Fixed #616 (originally discussed in #586) by updating the conditional visibility rules on the flexipage `LogEntryRecordPage`. Now, the 'HTTP Response' section will show up if any of these 3  conditions are `true`:
  - `HttpResponseStatus__c != null` (existing check)
  - `HasHttpResponseBody__c == true`
  - `HasHttpResponseHeaderKeys__c == true`
- Added the new-ish Analytics standard tab to the navigation items in the LoggerConsole app. This tab is generally available (GA) as of [Salesforce's Summer '23 release](https://help.salesforce.com/s/articleView?id=release-notes.rn_rd_unified_home_ga.htm&release=244&type=5).

  image

# Pipeline Changes
- Upgraded most of the `devDependencies` in `package.json`
- Hopefully fixed package generation for the managed package by including an empty copy of `Admin.profile-meta.xml` ಥ_ಥ
- Updated both `sfdx-project.json` files (1 for the unlocked package & 1 for the managed package) to use the new flag `"scopeProfiles": true`


jongpie:
This has been released in [`v4.12.8`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.12.8)

PR #631, opened 2024-02-08
Spring '24 Release (API v60.0) - v4.13.0

### LINKED ISSUE #533: How do I use the lwc logger with the Nebula namespace?
### New Feature Summary

I'm following the example in the readme and substituting `c` with `Nebula` in my lwc component, so the lwc contains

`<Nebula-logger></Nebula-logger>`

When I deploy this I get the error `No MODULE named Nebula:logger found`

What is the correct way to use this in an lwc?

### LINKED ISSUE #615: Logs are not generating each time for Lwc and Aura
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/581

<sup>Originally posted by **monika-gupta-erudite** November  7, 2023</sup>
I am exploring this logger for LWC and Aura 
Created a simple LWC to create logs on a button click
Lwc
Logs are not Generating on single click it gets generated on double click of the button.
and generated logEntry record does not have information about the LWC component  
image
Same thing happening with my Aura component .
LogEntry record generated from Aura component saying Component type as LWC.  
I tried changing the save methods also.</div>

### LINKED ISSUE #624: RelatedLogEntriesController_Tests sometimes throws an uncatchable UnexpectedException when using SOSL
### Discussed in https://github.com/jongpie/NebulaLogger/discussions/621

<sup>Originally posted by **gauravchaube21** February  1, 2024</sup>
We have installed Nebula Logger as an unlocked package in our org. Our pipeline installs the Nebula Logger to a scratch org whenever it gets created. Along with the package, we install the regular metadata and run test classes. This happens via our CI/CD Pipeline. We use the following coming to run the test classes
sf apex run test --test-level RunLocalTests --wait 30 --code-coverage --detailed-coverage --result-format human

Unfortunately, even though we have specified `RunLocalTests` in our sf command, all of the test classes run when running the command. Infact, a lot of test classes fail (Error Message: `Search Unavailable`) when this happens. 
image (2)

It would be great if someone could help us with the reason why the package test classes run even though we explicitly run local test classes and a possible workaround in this situation. Our pipelines are blocked because of the failing test classes within the package.

Thanks in advance</div>

jongpie:

## Spring '24 Release Upgrade
- Bumped all metadata to API v60.0 (Spring '24 release)
- Optimization & cleanup - Switched to using [the new standard class `System.UUID`](https://help.salesforce.com/s/articleView?id=release-notes.rn_apex_uuid.htm&release=248&type=5), instead of [my custom UUID class](https://github.com/jongpie/ApexUUID). The new system class is soooo much faster.

## `logger` LWC Bugfixes & Improvements + Enabled `logger` LWC in the Managed Package
- Worked with @jamessimone to fix #615 by optimizing how `logger` LWC loads user settings async (as well as several other optimizations throughout the LWC). The component should now log & save entries correctly, but developers need to either use `await` or chain `.then()` when calling `createLogger()`.
  
  Here is an example of using `await`

  ```javascript
  import { createLogger } from 'c/logger';
  
  export default class LoggerLWCDemo extends LightningElement {
    logger;
    
    async connectedCallback() {
      this.logger.info('Hello, world');
      this.logger.saveLog();
    }
  }
  ```
  
  This example uses `.then()` in `connectedCallback()` (instead of `await`). Additionally, this example shows `renderedCallback()` using a truthy check on `logger` to ensure it's been initialized before trying to use it.
  
  ```javascript
  import { createLogger } from 'c/logger';
  
  export default class LoggerLWCDemo extends LightningElement {
    logger;
    
    connectedCallback() {
      createLogger().then(async logger => {
        this.logger = logger;
        this.logger.info('Hello, world');
        await this.logger.saveLog();
      });
    }
  
   // and maybe even ...
    renderedCallback() {
      // this.logger may not be defined for initial renderings
      // you can either use a guard clause:
      if (this.logger) {
        this.logger.info('Some info about a re-rendering');
      }
      // or safe navigation:
      this.logger?.info('Some info about a re-rendering');
    }
  }
  ```
- Fixed #533 by updating `logger.js-meta.xml` so the `isExposed` node is set to `true` - this change makes the `logger` LWC available to use in the managed package for JavaScript logging. You can import it using the `Nebula` namespace:

  ```javascript
  import { createLogger } from 'Nebula/logger';
  
  export default class LoggerLWCDemo extends LightningElement {
    logger;
    
    async connectedCallback() {
      this.logger.info('Hello, world');
      this.logger.saveLog();
    }
  }
  ```

## Bugfixes 
- Fixed an issue in `LoggerDataStore` where not all string fields were truncated for platform events. This functionality was originally [introduced in v4.12.6](https://github.com/jongpie/NebulaLogger/releases/tag/v4.12.6)
  - Also optimized how the code looks for string fields to truncate to reduce some unnecessary looping
- Fixed an issue with the optional Status API integration (used to retrieve additional details about the current org's Salesforce instance) - Updated `LogManagementDataSelector` to always return null in `getCachedRecentLogWithApiReleaseDetails()` when `LoggerParameter.CALL_STATUS_API` is `false`
    - In previous versions, this method would still query the `Log__c` object, even if the optional Status API integration was disabled. But when the callout is disabled, this query will never be able to find a match, because the relevant fields haven't been populated by the callout. And in orgs with large data volume (LDV) of logging data, the condition `ApiReleaseNumber__c != NULL` is a very expensive query operation. 
    - This change will now skip the unnecessary (and sometimes expensive) SOQL query when the integration is disabled
- Fixed #624 by commenting out a flappy test in `RelatedLogEntriesController_Tests` that's been frequently failing due to ongoing platform issues with SOSL. It throws an uncatchable `UnexpectedException` due to Salesforce platform issues, resulting in (apparently) several pipelines failures for multiple companies that use Nebula Logger in their CI/CD jobs 😢 
    - "Fixed" ¯\\_(ツ)_/¯  - The functionality being tested has not changed in a few years, so commenting-out the test for now seems safe, and should prevent unnecessary pipeline failures for anyone that's been impacted by this problem.


nebula-logger/core/main/logger-engine/lwc/logger/loggerService.js
```diff
@@ -7,234 +7,130 @@ import { newLogEntry } from './logEntryBuilder';
...
 };
 
-/**
- * @return {LoggerService} a LoggerService instance
- */
-const createLoggerService = function () {
-    return new LoggerService();
+const createLoggerService = async function () {
+    const service = new LoggerService();
+    await service._loadSettingsFromServer();
```
jongpie, line 132: @jamessimone I was thinking about the function `_loadSettingsFromServer()` - right now, if there are multiple instances of `logger` on a page (e.g., multiple LWCs are on the same page, and each LWC uses `logger`), then each call to `createLoggerService()` will try to call the Apex method `getSettings()`. But the result of `getSettings()` will be the same for every call, since it returns settings for the current user.

The `getSettings()` Apex method is marked as `cacheable=true`, so maybe it's fine as-is. But the current version of the LWC stores [the settings object in a static variable](https://github.com/jongpie/NebulaLogger/blob/v4.12.9/nebula-logger/core/main/logger-engine/lwc/logger/loggerService.js#L18), which seems more effective. I tried reintroducing this into the current branch, but using a `static` variable breaks multiple jest tests, because the `static` variable isn't cleared between jest tests.

Any thoughts? Do you think we should leave it as-is? Or do you see a way to cache the result of the Apex method (without breaking jest tests)?

jamessimone: I'd let it be handled by the controller method being annotated with `cacheable=true` for now. I think that should suffice!

jongpie: Sounds good, I'll leave it as-is. Thanks!

PR #664, opened 2024-04-09
Added support for custom field mappings

### LINKED ISSUE #655: Add support for setting custom fields in LogEntryEventBuilder class
Hi Team,

I want to use managed packaged version of nebula logger. I want to know is there any plugin available
so that i can create my custom method in LogEntryEventBuilder class.

I am able to so that in unmanaged package and my expected functionality is working fine .

but i want to archive the same in managed package version of nebula so i need plugin information to explore the possibilities.

Pankaj

jongpie:
Resolves #655 by adding support for defining, setting, and mapping custom fields within Nebula Logger's data model. This is helpful in orgs that want to extend Nebula Logger's included data model by creating their own org/project-specific fields.

### Adding custom fields to the platform event `LogEntryEvent__e`
For orgs that only use `LogEntryEvent__e` (and don't leverage the custom objects), they can now...
- Create their own custom fields on `LogEntryEvent__e`. Any data type supported by platform events can be used.
  - In this example, a custom text field called `SomeCustomField__c` has been added:
    
    image

- Populate their own custom fields in Apex by calling the instance method overloads `LogEntryEventBuilder.setField(Schema.SObjectField field, Object fieldValue)` or `setField(Map<Schema.SObjectField, Object> fieldToValue)`.
  - For now, this functionality will only be available in Apex. Long-term, Flow & Lightning Components might also benefit from having equivalent functionality, but that will be handled separately.
  - Example: `Logger.info('hello, world').setField(LogEntryEvent__e.SomeCustomField__c, 'some value');`

### Adding custom fields to the custom objects `Log__c`, `LogEntry__c`, and `LoggerScenario__c`
For orgs that want to store the data in one of Nebula Logger's custom objects, they can follow the above steps, and also...
- Create their own custom fields on one of Nebula Logger's custom objects - right now, only `Log__c`, `LogEntry__c`, and `LoggerScenario__c` are supported.
  - In this example, a custom text field _also_ called `SomeCustomField__c` has been added to `Log__c` object - this will be used to store the value of the field `LogEntryEvent__e.SomeCustomField__c`:

    image

- Create a record in the new CMDT `LoggerFieldMapping__mdt` to map the `LogEntryEvent__e` custom field to the custom object's custom field, shown below. Nebula Logger will automatically populate the custom object's target field with the value of the source `LogEntryEvent__e` field.
  - In this example, a custom text field called `SomeCustomField__c` has been added to both `LogEntryEvent__e` and `Log__c`.

    image


nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -603,6 +603,50 @@ global with sharing class LogEntryEventBuilder {
...
+    }
+
+    /**
+     * @description Sets multiple field values on the builder's `LogEntryEvent__e` record
+     * @param  fieldToValue An instance of `Map<Schema.SObjectField, Object>` containing the
+     *                      the fields & values to populate the builder's `LogEntryEvent__e` record
+     * @return The same instance of `LogEntryEventBuilder`, useful for chaining methods
+     */
+    @SuppressWarnings('PMD.AvoidDebugStatements')
+    global LogEntryEventBuilder setFields(Map<Schema.SObjectField, Object> fieldToValue) {
```
jongpie, line 628: TODO decide if the method name should just be `set()` (or `put()`? 🤔 ) - the name would simpler & consistent for the overload/avoids the different names `setField()` and `setFields()`

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -603,6 +603,50 @@ global with sharing class LogEntryEventBuilder {
...
     }
 
+    /**
+     * @description Sets a field values on the builder's `LogEntryEvent__e` record
+     * @param  field      The `Schema.SObjectField` token of the field to populate
+     *                    on the builder's `LogEntryEvent__e` record
+     * @param  fieldValue The `Object` value to populate in the provided field
+     * @return The same instance of `LogEntryEventBuilder`, useful for chaining methods
+     */
+    global LogEntryEventBuilder setField(Schema.SObjectField field, Object fieldValue) {
```
jongpie, line 613: TODO decide if the method name should just be `set()` (or `put()`? 🤔 ) - the name would simpler & consistent for the overload/avoids the different names `setField()` and `setFields()`

nebula-logger/core/main/configuration/objects/LoggerFieldMapping__mdt/validationRules/UnsupportedTargetSObjectType.validationRule-meta.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<ValidationRule xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>UnsupportedTargetSObjectType</fullName>
+    <active>true</active>
+    <errorConditionFormula>AND(
+    TargetSObjectType__r.DeveloperName != &apos;Log&apos;,
+    TargetSObjectType__r.DeveloperName != &apos;LogEntry&apos;,
+    TargetSObjectType__r.DeveloperName != &apos;LoggerScenario&apos;
+)</errorConditionFormula>
```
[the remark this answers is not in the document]
jongpie: This seems to work, at least well enough for now - if you have both the unlocked & managed package installed, then you could unintentionally select the wrong package's objects/fields, so the validation should/could be a little stricter about validating the object's namespace, but there doesn't seem to be an easy way to setup the validation rule to work in both the managed package & unlocked package. So, this seems good enough for now.

nebula-logger/core/main/configuration/objects/LoggerFieldMapping__mdt/validationRules/UnsupportedSourceSObjectType.validationRule-meta.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<ValidationRule xmlns="http://soap.sforce.com/2006/04/metadata">
+    <fullName>UnsupportedSourceSObjectType</fullName>
+    <active>true</active>
+    <errorConditionFormula>SourceSObjectType__r.DeveloperName != &apos;LogEntryEvent&apos;</errorConditionFormula>
```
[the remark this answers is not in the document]
jongpie: This seems to work, at least well enough for now - if you have both the unlocked & managed package installed, then you could unintentionally select the wrong package's objects/fields, so the validation should/could be a little stricter about validating the object's namespace, but there doesn't seem to be an easy way to setup the validation rule to work in both the managed package & unlocked package. So, this seems good enough for now.

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
jongpie: @jamessimone if you happen to have some time to spare, I'd love get your thoughts on the naming convention for these 2 new builder methods. Right now, I'm using:
- `setField(Schema.SObjectField field, Object fieldValue)` for setting a single field
- `setFields(Map<Schema.SObjectField, Object> fieldToValue)` for setting multiple fields

The slight difference in method name (`setField` vs `setFields`) could be fine, but I was also considering calling by methods `set()`, like:
- `set(Schema.SObjectField field, Object fieldValue)` for setting a single field
- `set(Map<Schema.SObjectField, Object> fieldToValue)` for setting multiple fields

I can't decide if I prefer shorter & consistent overload name `set()` - I think the current method names (`setField` and `setFields`) are a bit more descriptive, but I'm not always a fan of method names being singular/pluralized based on the parameters ("field" vs "fields").

Let me know what you think!

jamessimone: I tend to prefer names that work for either argument, but `set` is part of the [reserved keywords](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/apex_reserved_words.htm) though I think it's one of those weird ones where it's not technically enforced.

For a package - particularly in this day and age, without necessarily having intellisense on package methods - I could see the `setField` / `setFields` nomenclature helping.

Either way, my guess is whatever you choose will be fine!

jongpie: Good point about `set` being a reserved word - it does indeed work (I'm able to deploy a class with methods named `set()`), but it seems safer to avoid using that as a method name. I thought about using `put()` instead, but almost all of the existing methods in `LogEntryEventBuilder` use the convention `setSomething()` (e.g., `setMessage()`, `setRecord()`, `setHttpRequestDetails()`, etc.), so the use of the word `put` feels a little inconsistent, even though that's the SObject method being used.

I think I'm leaning towards staying with `setField()` and `setFields()`, but I'll think about it a bit more this week before wrapping up this PR.

Thanks for your help!

jongpie: I thought about this a bunch, and I've decided to go with `setField()` for both methods (making them overloads). In the end, the deciding factor for me was `setField()` (singular) is consistent with other builder method overloads, like `setRecord()`, which has overloads for `SObject`, `List<SObject>`, and `Map<Id, SObject>`.

I still struggle with which convention I prefer in general - but I love consistency, so `setField()` seems like the most consistent option relative to the rest of Nebula Logger's codebase.

nebula-logger/core/main/configuration/objects/LoggerFieldMapping__mdt/fields/TargetField__c.field-meta.xml
```diff
@@ -0,0 +1,16 @@
...
+    <inlineHelpText
+    >The target field to populate with the value of the source field. The data type of this field must match the data type of the source field - mappings that use different/incompatible data types will be ignored, resulting in lost logging data.</inlineHelpText>
+    <label>Target Field</label>
+    <metadataRelationshipControllingField>LoggerFieldMapping__mdt.TargetSObjectType__c</metadataRelationshipControllingField>
+    <referenceTo>FieldDefinition</referenceTo>
+    <relationshipLabel>Logger Target Field Mappings</relationshipLabel>
+    <relationshipName>LoggerTargetFieldMappings</relationshipName>
+    <required>true</required>
+    <type>MetadataRelationship</type>
+    <unique>true</unique>
```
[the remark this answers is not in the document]
jongpie: After talking through this with @jamessimone some more, I'm going to stick with using `unique == true` here - I can't think of a valid situation where multiple mappings should exist for the same target field (since only one value ultimately can be saved to the field), so making this unique makes sense.

Issue #691, opened 2024-05-25
logger does not list component type when used in aura component


ZackFra:
### Package Edition of Nebula Logger

Unlocked Package

### Package Version of Nebula Logger

v4.13.11

### New Bug Summary

Attempted to use the lwc logger in an aura component. Retrieved the logger with the following code and saw that although the component section appeared on the log entry page, it was blank.

component file:
```
<aura:component implements="flexipage:availableForRecordHome" controller="ComponentLogger">
    <c:logger aura:id="logger"/>
    <lightning:button label="Log" onclick="{!c.log}"/>
</aura:component>
```
controller file
```
({
    log: function(component, event, helper) {
		const logger = component.find('logger');
		logger.info('Logging with info');
		logger.saveLog();
    }
})
```

Observed that the following log entry was created after clicking "Log".

![image](https://github.com/jongpie/NebulaLogger/assets/52726500/b8d8aafc-6f92-4c90-b777-e707973e4f1b)

The component section was visible, but not populated.

![image](https://github.com/jongpie/NebulaLogger/assets/52726500/59e9a72a-8da7-4736-961f-4bf1120e8a69)

Suspected this had something to do with the stack being generated in the LWC, but I would think that would've marked this log entry as an LWC if that was the case?

![image](https://github.com/jongpie/NebulaLogger/assets/52726500/48493ba2-c10b-4510-8ca5-76e0f3e1b5b8)

Played around with it for a few hours, modified the code a bit, injected the "stack" attribute with a stack generated from the aura component and passed down, still saw the same results.

ZackFra:
Worth noting, this issue was also raised today by another person on Reddit. They found that when importing the component directly, in an LWC, it worked correctly, but seemed to work as I've described when grabbing the injected logger component via a template query.

https://www.reddit.com/r/SalesforceDeveloper/comments/1czqs5t/nebula_logger_with_lwcs/

jongpie:
[...]

PR #692, opened 2024-05-26
bugfix/aura-logging

ZackFra:
Found that trying to log via aura components didn't format correctly. Raised an issue about it.

I expect this to get rejected, since I am aware that some of the Logger_Test methods are failing. Didn't want to mess with them without some sort of input. Looks like they might be using old stack traces that Salesforce doesn't generate anymore, that's my best guess anyway.

In a nutshell, found that aura components can't pull in the createLogger function that's exported by the logger lwc, and the stack trace captured by info, debug, etc. won't capture the aura component in the stack trace when you call them from component (i.e. component.find('logger').info('...'). 

Updated the logger LWC to have a createLogger function that exports `this`. After calling this new createLogger function, it works correctly.

There were issues with generating the stack trace in JS when used in a callback (i.e. createLogger().then( () => { ... }) so refacted createLogger to be synchronous, and to use JIT logic to pull in the settings. i.e. if the settings aren't loaded, saveLog will just await the this.#settingsPromise until it resolves.

Updated the LoggerStackTrace apex class to parse the stack traces from JavaScript with updated logic.


jongpie:
@ZackFra thanks for reporting this & submitting a PR! And apologies for the late response, I've had limited time lately to work on open source stuff, but hope to give this a more thorough review in the coming weeks.

Offhand, I have a couple of concerns with this PR - it looks like it changes/breaks logging in existing aura components & LWCs
- Aura components will now need to call `createLogger()`, which breaks existing implementations
- LWCs would potentially have issues with `createLogger()` no longer being async
- The change to `_meetsUserLoggingLevel(logEntryLoggingLevel)` (snippet below) will cause a bug in logging - by returning `true` when settings haven't been loaded, it will end up logging entries that shouldn't be logged. The same approach was used in Nebula Logger a while ago (last year?), and was changed to avoid this problem
```javascript
  _meetsUserLoggingLevel(logEntryLoggingLevel) {
    // return true until settings are loaded
    if(!this.#isSettingsLoaded) {
      return true;
    }
    ...
  }
```

I'll review + test this more ASAP!

jongpie:
I'm closing this PR, and will instead merge #727, which moves JS stack trace parsing to happen in JS directly (instead of in Apex). Some of the changes in this PR have been incorporated into #727, but since the entire architecture was changed, it was cleaner to create a new branch/PR for it.

Some related info for extra context & future reference:
- [My comment on issue #691](https://github.com/jongpie/NebulaLogger/issues/691#issuecomment-2272150030)
- The description in PR #727 gives more background info on some of the issues encountered + the thoughts behind changing the overall approach

PR #703, opened 2024-06-23
Implement relatedLogEntries refresh button + spinner

andrii-solokh:
**Motivation**
Working on new automation on a client project, I was generating logs in apex and attaching them to records. Each time I need to refresh the whole page, it was a bit annoying, so I thought it would be better to have a button for refresh or to auto refresh when new log created for currently open record.

**Solution**
- I'm using 'Change Data Capture' events to track as new Log Entry records arrived, then refreshing Lighting Data Service results. Initially implemented 'RefreshView API', so after record update lwc should reload related log entries, but log were created with delay 1-2 seconds and they were not present in save refresh, only on second refresh, so I decided to use CDC.

**Changes**
- Added spinner to let a user know - component is loading.
- Added refresh button

**Demo**
Changing account name generates log on record.

https://github.com/jongpie/NebulaLogger/assets/10558860/65697457-161b-47a1-b924-af82fcd44ce1


andrii-solokh-mnfd:
Hi, @jongpie. Thank you for response, I can't find docs telling 5 objects are the limit, only number of concurrent subscribes.

But, okay i will remove CDC part.

jongpie:
Hi @andrii-solokh - apologies, I meant to include a source for the 5 object limit. It's documented [here on the page 'Default Change Event Allocations'](https://developer.salesforce.com/docs/atlas.en-us.change_data_capture.meta/change_data_capture/cdc_allocations_default_allocations.htm)

image

You can also test out the limit in a scratch org/sandbox - if you try to add more than 5 objects to CDC, you'll get this error:

image

andrii-solokh:
@jongpie, thank you for the explanation, it looks like I should be more careful with using CDC and events in general.

jongpie (approved):
@andrii-solokh thanks so much for working on this, and for removing the CDC-related changes. I think what you built using CDC was very cool (and a great use case for CDC), but since Salesforce's platform limitations for it are so low, it's hard to rely on it sometimes.

Your changes look great - I've added a few extra commits to create the new package version, and I'll be releasing this today as `v4.13.12`.

jongpie:
@andrii-solokh this now [has been officially released as `v4.13.12`!](https://github.com/jongpie/NebulaLogger/releases/tag/v4.13.12) 🥳 

Thanks again for taking the time to work on this, I really appreciate the help!

andrii-solokh:
@jongpie it was tiny one, I hope to be more helpful in future 👍🏼

PR #727, opened 2024-08-06
Overhauled JavaScript stack trace parsing

### LINKED ISSUE #691: logger does not list component type when used in aura component

jongpie:
This is a big PR, so here's a _lot_ of context to explain the issues at hand & what's changing.

## Context: The current JavaScript stack trace parsing logic is veeeeery broken
Using the `c/logger` LWC (or `Nebula/logger` in the managed package), JavaScript developers can adding logging to their own Aura components & LWCs. And much like logging in Apex, when logging in JavaScript, Nebula Logger (tries to) use stack trace parsing to automatically track details about your components, such as the name & function in your component that added a log entry. These details are then stored in fields like `LogEntry__c.OriginLocation__c`, `LogEntry__c.OriginSourceApiName__c`, `LogEntry__c.OriginSourceActionName__c`, etc.

Thanks to some great feedback & contributions from @ZackFra and others, it turns out that there are some big problems with the current approach. This PR fixes _several_ issues related to JavaScript stack trace parsing, including:

- The current JS stack trace parsing does not work in all browsers. Most of the current logic for JS stack trace parsing is based on the format of Firefox's stack traces - but different browsers use different stack trace formats, so the parsing logic was wrong
- The current JS stack trace parsing does not (usually) work in Aura components, or LWCs that still embed `logger` in markup (instead of using the `import` approach in JS).
  - Fixes #691 issue reported by @ZackFra
  - Related reddit post: https://www.reddit.com/r/SalesforceDeveloper/comments/1czqs5t/nebula_logger_with_lwcs/

This PR will be used in place of PR #692 - after some extensive testing of the current release and changes in #692, I found that there some other related parsing issues with the current release's overall approach, and resolving everything while using the same overall approach/architecture would be fairly challenging. But credit & many thanks to @ZackFra for identifying the critical problems in the current code & how to resolve them. This PR leverages several of the same approaches @ZackFra used in #692.

One outstanding issue _not_ addressed by this PR:

- In some browsers, log entries added during a callback/when using `async` functions (example below) have some details missing from their stack traces. This is in part due to `createLogger()` being an `async` function
  - @ZackFra addressed this in PR #692, but the changes would result in some breaking changes for existing LWCs/Aura components that using logging
  - A separate issue has been logged to track this problem - #728. Hopefully, there will be a solution that can avoid any breaking changes for orgs that will be upgrading Nebula Logger
  - Discussion #657 seeeeems to be caused by this issue (more testing will be needed to confirm)

```javascript
export default class TestLWC extends LightningElement {
  async connectedCallback() {
    this.logger.info(
      "this entry will have some missing origin details in certain browsers"
    );
  }
}
```

## Old Approach: JS Stack Trace Parsing Occurred in Apex

Previously, JavaScript stack traces were parsed by the Apex classes `ComponentLogger` and `LoggerStackTrace`. One thought behind using this approach was doing it in JavaScript would require the user's browser to do more work, so offloading the parsing to Apex (instead of JS) would help minimize any performance impact of using the `logger` LWC.

But, building the parsing in Apex has the downside of making Nebula Logger the only project in the Salesforce ecosystem (that I'm aware of) that's trying to parse JavaScript stack traces using Apex. I have not been able to find any available projects/libraries/classes to provide this - so it has to be built, and maintained, just for Nebula Logger.

Before realizing the number of variations possible in JS stack traces, using Apex still seemed reasonable. But based on some of the info found by @ZackFra in issue #691 and PR #692, I did some extensive testing (some of which is detailed [in this comment](https://github.com/jongpie/NebulaLogger/issues/691#issuecomment-2227468194)), using a combination of:

- Different browsers (Firefox, Chrome, and even... MS Edge 🤮)
- Different methods of using the LWC `c/logger`
  - Importing `createLogger()` in another LWC's `.js` file
  - Embedding `<c-logger>` in another LWC's `.html` file
  - Embedding `<c:logger>` in an Aura component's `.cmp` file
- Different Salesforce contents/targets for LWCs (and the equivalent interfaces in Aura components):
  - `lightning__AppPage`
  - `lightning__FlowScreen`
  - `lightning__HomePage`
  - `lightning__RecordAction`
  - `lightning__RecordPage`
  - `lightning__Tab`
  - `lightning__UtilityBar`
  - `lightningCommunity__Page`

## Improved Recipes Metadata: Testing These Issues Was Too Difficult

It took a _lot_ of effort to even be able to test these scenarios/permutations, as Nebula Logger's repo didn't have sample metadata setup for most of these scenarios.

Previously, I was really only testing the recipes demo components via custom tabs / the target `lightning__Tab`. And when testing that target, in Firefox (my preferred browser), the stack trace parsing (built in Apex) worked. But in several other contexts, and in almost every other browser, the stack trace parsing logic resulted in inaccurate or missing data.

This led to me spending some time focusing on just making it easy for me to be able test logging in different targets, using multiple browsers. This includes:

- Improving existing `recipes` metadata so that the 3 demo components (listed below) can now be easily tested in multiple contexts, using the `Logger Recipes` app
  - `<c:loggerAuraEmbedDemo>` Aura component
  - `<c-loggerLWCEmbedDemo>` LWC
  - `<c-loggerLWCImportDemo>` LWC
- Improving the Experience Cloud metadata stored in `config` (used by the pipeline & during development)
  - **New(er) Lightning Web Runtime (LWR)**: Rebuilt the current Experience Cloud site to embed the recipes demo components in some pages to make them easy to test
  - **Old Aura Framework**: Created a second Experience Cloud site, using the older Aura framework template. This site makes it easy to test the same recipes demo components, using the older runtime, to ensure JS stack trace parsing works correctly

Now, the recipes app is setup to quickly test the 3 demo components in multiple `targets`

image

## New Approach: JS Stack Trace Parsing Now Occurs in JS

Once I was able to test JS logging in multiple targets & browsers, I was able to see some huge/critical differences in the stack traces generated in several scenarios. This led to 2 decisions:

1. The complexity of trying to fix the Apex approach quickly grew: @ZackFra introduced in PR #692 some changes that improved the Apex code that parses JS stack traces (and fixed the parsing in several situations). But for some situations/browsers still did not work correctly, so I spent some trying extend the logic to cover more scenarios. The complexity of this quickly escalated.

2. When parsing the JS stack trace in Apex, there wasn't a great way to show any of the parsed stack trace's data in JS. Ideally, I want JS logging to be able to show some useful summary in `console.log()` statements (when enabled) to help developers better debug their components. But to do this, the frontend would need to have the parsed stack trace data. This has always been an issue with using Apex to (async) parse the JS stack traces

As a result, I looked into options for parsing stack traces directly in JavaScript, with 2 hopes/goals in mind:
1. I really don't want to have to build/maintain an Apex implementation that accommodates JS stack traces across all browsers. It seems hard to maintain 😅
2. I want to improve the output shown in `console.log()` statements to provide Salesforce developers with more context.

Eventually, I came across [the open source library JavaScript library `stacktrace.js`](https://github.com/stacktracejs/stacktrace.js). It provides cross-browser stack trace parsing, directly in JavaScript. After prototyping this out for several days, it seems to provide (with a few tweaks) accurate stack trace parsing for all of the permutations of browsers + Salesforce targets + methods of calling the `logger` LWC.

This PR incorporates a modified version of its stack trace parsing code - all of the Apex code that previously handled JS stack traces has now be removed (with a few lingering items that are deprecated & will be removed in a future release).

## Improved the Output of `console.log()` Statements

Lastly, since JS stack trace parsing now happens directly in JS, this makes it easier to provide some helpful context to the `console.log()` statements that Nebula Logger automatically executes. 

### Old Approach

In previous versions, Nebula Logger would automatically include the JS object representation of a component log entry every time it auto-calls a `console` log function. This is/was done to try to make it easy for JavaScript developers to see some logging context directly in their browser when debugging.

But due to how locker service/locker web security (LWC) work, the object was always shown as a `Proxy` object

image

To then actually see the object's data, JS developers would have to store the `Proxy` as a variable in their browser's dev console, and then run `JSON.parse(JSON.stringify(theProxyObject))`. So although the data was available in the browser, it was very tedious to actually leverage the data.

I also tried printing the `Proxy` object, using `JSON.stringify(theProxyObject)` - but because it contained the original, unparsed stack trace (and not a parsed version of it), the printed string was huuuuuge, and it was still very difficult to quickly see the relevant lines from the stack trace.

### New Approach

With JS stack trace parsing now happening in the browser, each JS logging statement now includes a pretty-printed JSON string that includes the most relevant info about the log entry's origin:
- The LWC/Aura component's name
- The function that generated the log entry
- The metadata type of your component (`AuraDefinitionBundle` or `LightningComponentBundle`)
- The timestamp of the log entry

Now developers can see at a glance the relevant info about their component that logged something.

Console output in Chrome:

image

Console output in Firefox:

image

Console output in Microsoft Edge:

image


nebula-logger/core/main/logger-engine/classes/ComponentLogger.cls
```diff
@@ -319,16 +361,40 @@ public inherited sharing class ComponentLogger {
...
+        public String componentName { get; set; }
+
+        /**
+         * @description The component's function that generated the log entry
+         */
+        @AuraEnabled
+        public String functionName { get; set; }
+
+        /**
+         * @description The metadata type of the component thatthat generated the log entry
```
jamessimone, line 387: thatthat

nebula-logger/core/main/logger-engine/classes/LoggerStackTrace.cls
```diff
@@ -31,15 +30,19 @@ public without sharing class LoggerStackTrace {
     @SuppressWarnings('PMD.FieldNamingConventions')
     public enum SourceLanguage {
         Apex,
+        // JavaScript is now deprecated (possible SourceLanguage altogether?),
```
jamessimone, line 33: possiby

nebula-logger/core/main/logger-engine/classes/LoggerStackTrace.cls
```diff
@@ -166,84 +166,6 @@ public without sharing class LoggerStackTrace {
         this.Location = cleanedLocation;
     }
```
jamessimone, line 168: nice! this is going to be a big performance improvement, the stacktrace parsing was still one of the big errors where bulk log entries were slowing down!

nebula-logger/core/main/logger-engine/lwc/logger/logEntryBuilder.js
```diff
@@ -180,11 +212,55 @@ const LogEntryBuilder = class {
...
+                    {
+                        origin: {
+                            component: this.#componentLogEntry.originStackTrace?.componentName,
+                            function: this.#componentLogEntry.originStackTrace?.functionName,
+                            metadataType: this.#componentLogEntry.originStackTrace?.metadataType
+                        },
+                        scenario: this.#componentLogEntry.scenario,
+                        timestamp: this.#componentLogEntry.timestamp
+                    },
+                    (_key, value) => value ?? undefined,
```
jamessimone, line 235: I think it's more common to use `_` as the unused variable name

nebula-logger/core/main/logger-engine/lwc/logger/loggerStackTrace.js
```diff
@@ -0,0 +1,290 @@
...
+      https://github.com/stacktracejs/stacktrace.js/blob/710bba1118d396466ee342a30b3dfd19ecbda8b5/dist/stacktrace.js#L14C9-L202C7
+
+Some of the changes made below for Nebula Logger's usage includes:
+    - converting the function ErrorStackParser(StackFrame) into a class
+    - converting var variables to const & let
+    - updating a few others lines to use shorter/newer JS syntax
+    - removal of the StackFrame class originally returned, and replaced with a simple Object
+    - updated some parsing logic to reflect the current structure of some JavaScript stack traces
+      (e.g., Chrome & Edge stack traces include 'Proxy.' before the function name, so 'Proxy.' is now stripped out)
+*/
```
jamessimone, line 24: awesome comment

jamessimone (approved):
@jongpie a couple of small comments, otherwise looks great!

nebula-logger/core/main/logger-engine/classes/LoggerStackTrace.cls
jongpie: Yeah, it took a while for me to come to terms with shifting this logic to JS, but I think in the end, getting rid of this Apex code is going to be great 🏎️

And I'm hoping to revisit 1 or 2 other parts of the remaining code (for parsing Apex stack traces) in the coming weeks - it's a bit out of scope for this PR, but I think I might be able to optimize some of that logic a bit to reduce CPU time some more.

jamessimone:
Awesome! Nice work

PR #770, opened 2024-09-25
Bugfix: Updated Logger.setAsyncContext() behavior

jongpie:
- Updated the behavior of `Logger.setAsyncContext()` to only set the context the first time a non-null context value is provided
  - Previously, subsequent calls would overwrite the context value, which wasn't really the intended behavior - it's intended to be used to set context at the beginning of a transaction

# Pipeline Changes
- Made some optimizations in `build.yml` so some steps don't run on draft PRs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,16 +8,20 @@ env:
 
 on:
   push:
+    branches:
+      - main
     paths:
       - .github/workflows/**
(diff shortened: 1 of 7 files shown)


jamessimone:
Aw yeah!

jongpie:
@jamessimone thanks again for reviewing this!

PR #772, opened 2024-09-25
Added new Apex method & JS function Logger.setField() 

### LINKED ISSUE #769: Add the ability to set custom fields once per transaction/on every LogEntryEvent__e record
This idea came up [today on SFXD](https://discord.com/channels/246568944213819393/348054624256786434/1286764962072629328) with @surajp. In [release `v4.13.14`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.13.14), a new instance method was added - `LogEntryEventBuilder.setField()` - to set custom fields on `LogEntryEvent__e`: 
- This is useful for fields that are entry-specific (i.e., the field ultimately is mapped to a custom field on `LogEntry__c`)
- This is... confusing for fields that are transaction specific (i.e., the field value ultimately is mapped to a custom field on `Log__c`). 
    - Right now, setting the custom field on every entry would probably be the best way to ensure the `Log__c` field is correctly populated. But doing it on _every_ entry in a transaction is very tedious / not a great developer experience

Ideally, developers should only have to set the transaction-level fields once per transaction. This should be handled with 2 changes:
1. **New static method `Logger.setField(SObjectField field, Object value)`**: This will give Apex developers a method that can be called once per transaction for a field, and all generated `LogEntryEvent__e` records will have it set
2. **Improve `Log__c` mappings in `LogEntryEventHandler`**: Right now, the trigger handler creates a `Log__c` record the first time it finds a `LogEntryEvent__e` record with a specific transaction ID - and then skips the log-related code for any subsequent events with the same transaction ID. In some situations, some `Log__c` data could be loss, so I'd like to improve how it works
    - If `event A` has `SomeLogField__c = null`, and `event B` has `SomeLogField__c = 'some value'`, then the generated `Log__c` could have a `null` value, if both events are published/processed in the same async transaction

jongpie:
## New Apex Method `Logger.setField()`
Resolved #769 by adding a new static method, `Logger.setField()`. This gives Apex developers a way to set field(s) once per transaction, and the fields will be auto-populated on all `LogEntryEvent__e` records generated in that transaction.
- The `setField()` method has 2 overloads (same as the instance method overloads `LogEntryEventBuilder.setField()` that were introduced in [release `v4.13.14`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.13.14))
  1. `Logger.setField(Schema.SObjectField field, Object fieldValue)` - useful for easily setting the value for a single field
  2. `Logger.setField(Map<Schema.SObjectField, Object> fieldToValue)` - useful for setting the value for multiple fields
- The new method supplements the functionality introduced in [release `v4.13.14`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.13.14), as shown below:

  ```apex
  // 🥳New!
  // Set a custom field on all LogEntryEvent__e records in a transaction,
  // using static method Logger.setField()
  Logger.setField(LogEntryEvent__e.My_Custom_Log_Field__c, 'the value for some Log__c field')

  // ℹ️Existing - introduced in v4.13.14
  // Set a custom field on a single LogEntryEvent__e record,
  // using instance method LogEntryEventBuilder.setField()
  Logger.info('hello, world')
    .setField(LogEntryEvent__e.Some_Custom_Log_Entry_Field__c, 'the value for some LogEntry__c field');
  ```

## New JavaScript Function `logger.setField()`
Added JavaScript support for setting fields once per component instance, using `logger.setField()`. This is the JS equivalent of the new Apex method `Logger.setField()` (above).

  ```javascript
  export default class LoggerLWCImportDemo extends LightningElement {
    logger = getLogger();

    connectedCallback() {
      // 🥳New!
      // Set My_Field__c on every log entry event created in this component with the same value
      this.logger.setField({My_Field__c, 'some value that applies to any subsequent entry'});

        // ℹ️Existing - introduced in v4.14.6
      this.logger.warn('hello, world - "a value" set for Some_Other_Field__c').setField({ Some_Other_Field__c: 'a value' });
    }
  }
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,13 @@ The most robust observability solution for Salesforce experts. Built 100% native
 
 ## Unlocked Package - v4.14.14
 
-[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015oW3QAI)
-[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015oW3QAI)
+[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015oWIQAY)
+[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015oWIQAY)
 [![View Documentation](./images/btn-view-documentation.png)](https://github.com/jongpie/NebulaLogger/wiki)
 
-`sf package install --wait 20 --security-type AdminsOnly --package 04t5Y0000015oW3QAI`
+`sf package install --wait 20 --security-type AdminsOnly --package 04t5Y0000015oWIQAY`
 
-`sfdx force:package:install --wait 20 --securitytype AdminsOnly --package 04t5Y0000015oW3QAI`
+`sfdx force:package:install --wait 20 --securitytype AdminsOnly --package 04t5Y0000015oWIQAY`
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 The most robust observability solution for Salesforce experts. Built 100% natively on the platform, and designed to work seamlessly with Apex, Lightning Components, Flow, OmniStudio, and integrations.
 
-## Unlocked Package - v4.14.13
+## Unlocked Package - v4.14.14
 
 [![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015oW3QAI)
 [![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015oW3QAI)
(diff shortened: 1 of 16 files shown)


jongpie:
@surajp I think the changes in this PR will handle the limitation that we discussed a few days ago with custom field mappings (issue #769). If you happen to have a some time to take a look at this, I'd love to hear your thoughts on the changes before I merge it.

surajp:
> @surajp I think the changes in this PR will handle the limitation that we discussed a few days ago with custom field mappings (issue #769). If you happen to have a some time to take a look at this, I'd love to hear your thoughts on the changes before I merge it.

That was fast and the changes look great! If I understand correctly, if I do `Logger.setField(LogEntryEvent__e.My_Custom_Field__c,'value1')` and add then `logger.info('some message').setField(LogEntryEvent__e.My_Custom_Field__c,'value 2')`, the `log__c` record will have `My_Custom_Field__c` set to `value 2`?

jongpie:
@surajp at the moment, that scenario would have slightly different behavior, depending on when you use the `setField()` methods, and where the `LogEntryEvent__e` field is mapped

### Example 1:
```apex
Logger.setField(LogEntryEvent__e.My_Custom_Field__c,'value1');
// This entry will have 'value1'
Logger.info('an entry that does not use the builder setField() method);
// This record will have 'value2'
Logger.info('some message').setField(LogEntryEvent__e.My_Custom_Field__c,'value 2');
```

In this case, the first entry would have `value1` - and essentially, the first `LogEntryEvent__e` for a particular transaction ID is used to create the `Log__c` record. So if `LogEntryEvent__e.My_Custom_Field__c` is mapped to a `Log__c` field, then the `Log__c` would have `value1`.

### Example 2
```apex
Logger.setField(LogEntryEvent__e.My_Custom_Field__c,'value1');
// This entry will have 'value2'
Logger.info('some entry that overrides the same field set via static method Logger.setField()').setField(LogEntryEvent__e.My_Custom_Field__c,'value 2');
```

In this case, since the first entry would have `value2` - and then the `Log__c` field would also have `value2`

I think this behavior makes sense, but let me know what you think about it. At the very least though, I'll want to document the behavior so people understand how it works in these situations.

jongpie:
@surajp I went ahead and released these changesin [`v4.14.14`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.14.14)! I think the changes in this PR will help improve the functionality a good bit, and I'll continue to look into some of the other enhancements/changes we discussed for custom fields/field mappings. Thanks again for your help & feedback!

PR #775, opened 2024-09-28
New synchronous JS function getLogger() + deprecated async function createLogger()

### LINKED ISSUE #728: JavaScript stack traces can still be inaccurate due to createLogger() being async
For some browsers + contexts, JS stack traces are missing info need to determine a JS log entry's origin details. This includes using `async connectedCallback()`
- Discussion #657 is possibly related/caused by this issue
- @ZackFra solved this issue in #692 by making `createLogger()` no longer async - but some of the related changes could have an impact on existing orgs, with code that _does_ expect it to be `async`. 

I'd like to explore 2 options:
  1. Find a way to make `createLogger()` no longer `async` without causing issues in existing code
  2. Keep `createLogger()` async, but introduce another function that isn't async - `createLogger()` would then be considered deprecated, but still supported for now

### LINKED ISSUE #763: Add new exception() function to logger LWC for JavaScript developers
Similar to the Apex method `Logger.exception()` that was introduced in [release `v4.11.2`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.11.2), there should be a JavaScript function that automatically logs a JS error, saves the log, and rethrows the error.

### LINKED ISSUE #776: Aura/LWC stack trace parsing is not accurate when lightning debug mode is disabled

Unlocked Package

4.14.12

### New Bug Summary

Hi @jongpie @ZackFra ,

I installed 4.14.12 version unlocked package , when i try to log for LWC ,the stack trace is not clear still the same issue persists for me what @ZackFra attached image ,i am using chrome .

This is the below code i am using

          <template>
            <c-logger></c-logger>
            <div>Nebule Logger Component</div>
            <lightning-button variant="base" label="click To Log" title="" onclick={log} class="slds-m-left_x-small">      
            </lightning-button>
           </template>
JS
log() {

    const logger = this.template.querySelector('c-logger');
    
     logger.error('ERROR!!!').addTag('TAG');
    
    logger.warn('WARN!!!');
    
    logger.info('INFO!!!');
    
    logger.debug('DEBUG!!!');
    
    logger.fine('FINE!!!');
    
    logger.finer('FINER!!!');
    
    logger.finest('FINEST!!!');
    
    logger.saveLog();
    
    }
Can you please help us with that

jongpie:
## New `getLogger()` function

Partially fixed #728 by adding a new function `getLogger()` in `logger` LWC that can be called synchronously, and deprecated the async function `createLogger()`
- This simplifies how developers use the `logger` LWC, and avoids some lingering JavaScript (JS) stack trace issues that occur in `async` functions
- The function `createLogger()` is still supported & functional, but it's now considered deprecated since it requires using `await` (which is no longer necessary with `getLogger()`)

> [!NOTE]
> This does not truly "fix" or improve the stack trace parsing used in `async` functions, and `async` functions will continue to have inaccurate function names reported in some browsers (typically, `eval` is listed as the function name). This is a ongoing challenge in JavaScript due to what info is/isn't available in stack traces in some browsers (most notably, webkit browsers like Chrome & Edge). 
>
> But with the new `getLogger()` function, Nebula Logger no longer _requires_ the use of `async` / `await` - and often the only reason developers were using `async` / `await` was to be able to use Nebula Logger. In these situations, developers can eliminate `async` / `await`, and the resulting log entries will have accurate JS function names.

### Example: Old `createLogger()` Usage

Previously, JS developers had to use `async`, `await`, and truthy checks to ensure that Nebula Logger's LWC had finished being loaded. This resulted in more complexity for developers, who just want to be able to log some stuff 😢 

```javascript
import { createLogger } from 'c/logger';

export default class LoggerDemo extends LightningElement {
  logger;

  // Some lifecycle event (typically connectedCallback()) has to run async & await createLogger()
  async connectedCallback() {
    this.logger.info('Hello, world');
    this.logger.saveLog();
  }

  // @wire functions run around the same time as connectedCallback(), but there's no guaranteed order
  // This result in some logging requiring truthy checks using the safe navigator "?.", and some loss of logging data could occur
  @wire(returnSomeString)
  wiredReturnSomeString({ error, data }) {
    this.logger?.info('>>> logging inside a wire function, if the logger is loaded');
    if (data) {
      this.logger?.info('>>> wire function return value: ' + data);
    }
    if (error) {
      this.logger?.error('>>> wire function error: ' + JSON.stringify(error));
    }
  }
}
```

### Example: New `getLogger()` Usage

Now, `await` is no longer needed, and a `logger` instance can be immediately initialized & used. This results in less code, and happier developers 🥳 

```javascript
import { getLogger } from 'c/logger';

export default class LoggerDemo extends LightningElement {
  logger = getLogger();

  connectedCallback() {
    // Immediately use this.logger - no await, no truthy checks
    this.logger.info('Hello, world');
    this.logger.saveLog();
  }
}

  @wire(returnSomeString)
  wiredReturnSomeString({ error, data }) {
    // Immediately use this.logger - no await, no truthy checks
    this.logger.info('>>> logging inside a wire function');
    if (data) {
      this.logger.info('>>> wire function return value: ' + data);
    }
    if (error) {
      this.logger.error('>>> wire function error: ' + JSON.stringify(error));
    }
  }
}
```

## New `exception()` function

Resolved #763 by adding a JS function equivalent to the Apex method `Logger.exception()`. Both of these do 3 things in 1 line of code:
1. Log an exception
2. Save the log
3. Rethrow the exception

Previously in JS, this would have been 3 lines of code:

```javascript
this.logger.error('An error occurred').setExceptionDetails(someJSError);
this.logger.saveLog();
throw someJSError;
```

Now, 1 line of code provides the same functionality:

```javascript
this.logger.exception('An error occurred', someJSError);
```

## More Improvements for JS Stack Trace Parsing

Fixed #776 by updating logic in `loggerStackTrace.js` to better handle parsing when [lightning debug mode](https://help.salesforce.com/s/articleView?id=sf.aura_debug_mode.htm&language=en_US&type=5) is disabled
Previously, stack traces worked when debug mode was enabled, but was still inaccurate in some browsers when debug mode was off due to some subtle differences in the generated stack traces.


jongpie:
@jamessimone I think this PR is ready for review - since we chatted earlier, I finished up those Jest tests we were looking at (for calling the various `console` functions), and ended up moving some of the code from `logEntryEventBuilder.js` to `loggerService.js`
- `logEntryEventBuilder.js` is now much simpler - it's really just a DTO class (`ComponentLogEntry`) and some builder methods, it has a lot less logic/complexity now
- `loggerService.js` now handles capturing browser context, calling `console` functions, and calling `lightning/logger`
- `logger.test.js` is a bit of a mess right now - it now contains similar test functions for all 3 supported approaches (listed below). I might eventually split this into separate test files later
  1. New `getLogger()` import 
  2. Now-deprecated `createLogger()` import
  3. Mostly-deprecated markup embed approach (this is still used for Aura though)

Let me know if/when you have a chance to take a look!

jamessimone:
> * `logger.test.js` is a bit of a mess right now - it now contains similar test functions for all 3 supported approaches (listed below). I might eventually split this into separate test files later.

One possibility that might help to clean things up in that test class such that you don't need to repeat yourself so much would be using parameterized tests:

```js
const getMarkupLogger = async () => {
  const logger = createElement('c-logger', { is: Logger });
  document.body.appendChild(logger);
  await flushPromises();
  return logger;
};

// and then in a describe block
  it.each([[createLogger], [getLogger], [getMarkupLogger]])('returns user settings', async loggingFunction => {
    getSettings.mockResolvedValue({ ...MOCK_GET_SETTINGS });
    const logger = await loggingFunction();

    const userSettings = logger.getUserSettings();

    expect(userSettings.defaultSaveMethod).toEqual('EVENT_BUS');
    expect(userSettings.isEnabled).toEqual(true);
    expect(userSettings.isConsoleLoggingEnabled).toEqual(false);
    expect(userSettings.isLightningLoggerEnabled).toEqual(false);
  });
``` 

In this example, each inner list passed to `.each` can have any number of arguments that are then passed to the second function, a la:

```js
 it.each([
    [createLogger, 'createLogger'],
    [getLogger, 'getLogger deprecated'],
    [getMarkupLogger, 'markup logger']
  ])('returns user settings', async (loggingFunction, text) => {
 // ...
})
```

What do you think? I will continue reviewing, but that might really help to cut down on the duplication in that file

README.md
```diff
@@ -432,21 +431,31 @@ Each `LogEntry__c` record automatically stores the component's type ('Aura' or '
...
 ```javascript
 // For LWC, import logger's createLogger() function into your component
-import { createLogger } from 'c/logger';
+import { getLogger } from 'c/logger';
+import callSomeApexMethod from '@salesforce/apex/LoggerLWCDemoController.callSomeApexMethod';
 
-export default class LoggerLWCImportDemo extends LightningElement {
-  logger;
+export default class LoggerDemo extends LightningElement {
+  // Call getLogger() once per component
```
jamessimone, line 442: What about something like this for `loggerService.js`:

```js
export function getLoggerService() {
  return !LoggerService.loggerService ? new LoggerService() : LoggerService.loggerService;
}

export default class LoggerService {
  static hasInitialized = false;
  static loggerService;
  // etc...
    constructor() {
    this._loadSettingsFromServer();

    if (areSystemMessagesEnabled && !LoggerService.hasInitialized) {
      this._logToConsole('INFO', 'logger component initialized\n' + JSON.stringify(new BrowserContext(), null, 2));

      LoggerService.hasInitialized = true;
      LoggerService.loggerService = this;
    } else if (!LoggerService.hasInitialized) {
      LoggerService.hasInitialized = true;
      LoggerService.loggerService = this;
    }
  }

   // etc ...
}
```

And then in `logger.js`, just swap out the import and call that function:

```js
export function getLogger() {
  return getLoggerService();
}
```

That way it doesn't matter how many times `getLogger()` is called

nebula-logger/core/main/logger-engine/lwc/logger/loggerService.js
```diff
@@ -66,71 +121,125 @@ const LoggerService = class {
...
+      default:
+        consoleLoggingFunction = console.debug;
+        break;
+    }
+
+    const loggingLevelEmoji = LOGGING_LEVEL_EMOJIS[loggingLevel];
+    const qualifiedMessage = `${loggingLevelEmoji} ${loggingLevel}: ${message}`;
+    // Some JS stack traces are huuuuge, so don't print it in the browser console.
+    // The stack trace will still be saved on the backend.
+    // '\n' + JSON.stringify(this.#componentLogEntry, replacer, 2)
```
jamessimone, line 225: this looks like it can be removed

nebula-logger/core/main/logger-engine/lwc/logger/loggerService.js
```diff
@@ -66,71 +121,125 @@ const LoggerService = class {
...
-};
+  _meetsUserLoggingLevel(logEntryLoggingLevel) {
+    return this.#settings.isEnabled === true && this.#settings.userLoggingLevel.ordinal <= this.#settings.supportedLoggingLevels[logEntryLoggingLevel];
+  }
 
-export { createLoggerService };
+  /* eslint-disable no-console */
+  _logToConsole(loggingLevel, message, componentLogEntry) {
+    let consoleLoggingFunction;
+    switch (loggingLevel) {
```
jamessimone, line 206: what if instead of the switch statement here you used:

```js
let consoleLoggingFunction = console[loggingLevel.toLowerCase()];
if (!consoleLoggingFunction) {
  consoleLoggingFunction = console.debug;
}
```

nebula-logger/core/main/logger-engine/lwc/logger/loggerServiceTaskQueue.js
```diff
@@ -0,0 +1,59 @@
...
+
+    /* eslint-disable no-await-in-loop */
+    while (this.#taskQueue.length > 0) {
+      const task = this.#taskQueue.shift();
+      task.output = task.actionFunction(...task.actionArguments);
+      if (task.output instanceof Promise) {
+        task.output = await task.output;
+      }
+      processedTasks.push(task);
+    }
```
jamessimone, line 53: if you want, you can slightly simplify this to:

```js
    /* eslint-disable no-await-in-loop */
    this.#taskQueue.forEach(async task => {
      task.output = await task.actionFunction(...task.actionArguments);
      processedTasks.push(task);
    });
    this.#taskQueue = [];
```

jamessimone (changes requested):
@jongpie I left a few comments that I'd be curious to hear your thoughts on, especially the one for cleaning up `logger.test.js`

Overall this is looking good! Hopefully we can chat about all of this today

jongpie:
> > * `logger.test.js` is a bit of a mess right now - it now contains similar test functions for all 3 supported approaches (listed below). I might eventually split this into separate test files later.
> 
> One possibility that might help to clean things up in that test class such that you don't need to repeat yourself so much would be using parameterized tests:
> 
> ```js
> const getMarkupLogger = async () => {
>   const logger = createElement('c-logger', { is: Logger });
>   document.body.appendChild(logger);
>   await flushPromises();
>   return logger;
> };
> 
> // and then in a describe block
>   it.each([[createLogger], [getLogger], [getMarkupLogger]])('returns user settings', async loggingFunction => {
>     getSettings.mockResolvedValue({ ...MOCK_GET_SETTINGS });
>     const logger = await loggingFunction();
> 
>     const userSettings = logger.getUserSettings();
> 
>     expect(userSettings.defaultSaveMethod).toEqual('EVENT_BUS');
>     expect(userSettings.isEnabled).toEqual(true);
>     expect(userSettings.isConsoleLoggingEnabled).toEqual(false);
>     expect(userSettings.isLightningLoggerEnabled).toEqual(false);
>   });
> ```
> 
> In this example, each inner list passed to `.each` can have any number of arguments that are then passed to the second function, a la:
> 
> ```js
>  it.each([
>     [createLogger, 'createLogger'],
>     [getLogger, 'getLogger deprecated'],
>     [getMarkupLogger, 'markup logger']
>   ])('returns user settings', async (loggingFunction, text) => {
>  // ...
> })
> ```
> 
> What do you think? I will continue reviewing, but that might really help to cut down on the duplication in that file

👀 Ooooh, yeah, I _really_ like this idea, it's beautiful - I think this sounds like a much better approach. Before this PR, it already felt bad having two versions of the tests (one for `createLogger()` import and one for the old markup approach) - and now that this PR adds a third version, it's definitely been challenging to try to add consistent tests for all 3 approaches. In fact, I just gave up on consistency towards the end of the changes in this PR - the new tests I added for `console` function calls were only added for the `getLogger()` approach, even though I've typically been adding equivalent (but nearly identical) tests for the 2 other approaches. But it was becoming very tedious, so I think the current approach of duplication is... not fun. `F--`, would not recommend 👎 

With all that said, I might hold off on making this change right now, and instead do it in another PR in the coming weeks. There are several things related to Jest tests that I'd like to revisit:
- I've really let the LWC test coverage for the whole repo slip over the last few months. ~Despite code coverage being a _terrible_ thing for companies to use as a metric,~ I know in this case that the drop is because I haven't been putting as much effort into some of the Jest tests as I should be, and there's some functionality that I'd like to ensure is tested properly
- Some of the Jest tests haven't been updated in a while, so there's still some old test patterns being used that I'd like to cleanup (I was so young & naive with how I did some of the older tests 😅 ).
- I'd like to also have all the Jest tests run twice - once with a namespace prefix, and once without. A couple of the LWCs do test this already (e.g., `c/logEntrEventStream`), but the approaches used aren't consistent, and not all of the LWCs do it (e.g., `c/loggerSettings` doesn't, but should since `LoggerSettings__c` and its fields have a namespace prefix in the managed package).
  - Using `logger.test.js` as an example: I feel like there's probably some way to combine your suggestion of `it.each()` to have it tests all 3 approaches, but do it twice - once with a namespace, and once without (6 total variations of the tests would run)

I think I'll make another issue to cover all of the Jest stuff to improve in the repo (including using your approach above with `it.each()` in `logger.test.js`). Let me know if that approach makes sense

jamessimone:
That makes perfect sense!

nebula-logger/core/main/logger-engine/lwc/logger/loggerService.js
jongpie: ...but wait, what about this one-liner version??

```javascript
const consoleLoggingFunction = console[loggingLevel.toLowerCase()] ?? console.debug;
```

nebula-logger/core/main/logger-engine/lwc/logger/loggerServiceTaskQueue.js
jamessimone: I think you're right in saying that `forEach()` would _not_ iterate over a newly added task, but I guess I thought the `#isProcessing` check would already prevent something from being added to the queue. Maybe a test for this is in order?

README.md
jongpie: I think at least for now, it shouldn't return a static instance of the service - I'm definitely open to changing some of this, but using static would ~break~ change the existing behavior. Since JavaScript/lightning components don't have a direct equivalent to a transaction in Apex/Flow, it's still helpful to have some kind of "boundary" in JS to group related log entries - so each component having a separate instance sort of serves as the "boundary", each `LoggerService` instance manages all of the entries for a particular component.

- Each instance of `LoggerService` has its own buffer of log entries, and when they're saved, they're all related to the same `Log__c` record. This makes it easier to find log entries created from `c/someComponent` vs `c/anotherComponent`
- With a static instance of `LoggerService`, any time 1 component calls `saveLog()`, log entries from all LWCs on the page would get saved (which could be nice) & then tied to a single `Log__c` (which changes reporting/could be confusing...?)

I definitely want to consider changing/improving how some of this works in JS (and there's a semi-related open issue #496), but changing it will impact some orgs (in terms of reporting, etc), so I want to hold off on doing so for the moment. But let's definitely chat about this more whenever we have a chance.

jongpie: Hmmm... at least for now, I could make the `#settings` variable in `LoggerService` a static variable - no matter how many instances of `LoggerService` there are at any point, the settings should be the same, so making it static makes sense, I think. That should cut that Apex controller call down to 1 call per page load (instead of 1 call per instance of `LoggerService`)

nebula-logger/core/main/logger-engine/lwc/logger/loggerService.js
jongpie: I've_Looked_At_This_For_Five_Hours_Now_Banner

PR #787, opened 2024-10-23
Slightly tweaks async context to account for additional finalizer fields

jamessimone:
@jongpie wanted to get your thoughts on these changes:

* potentially also creating a log entry for async context when it's first set (possibly _always_ logging? or also reporting on when it's been called with the current context _already_ set?)
* appending some additional properties to the `Logger.AsyncContext` inner class to capture more data about finalizers. This leads to an INFO message like such:

```
INFO|FinalizerExample.execute
Nebula Logger - Async Context: {
  "type" : "System.FinalizerContext",
  "parentJobId" : "7076g0000B6tiGgAQI",
  "finalizerResult" : "UNHANDLED_EXCEPTION",
  "finalizerException" : "Oops!\nExternal entry point"
}
```

Not terribly pleased with the existing "stacktrace" string there (as far as I can tell, at the moment it always seems to be reported as an external entry point) but it might be good to leave in (in the hopes that we can get that worked on internally).

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,41 +238,41 @@ This example batchable class shows how you can leverage this feature to relate a
 
 ```apex
 public with sharing class BatchableLoggerExample implements Database.Batchable<SObject>, Database.Stateful {
-    private String originalTransactionId;
+  private String originalTransactionId;
 
-    public Database.QueryLocator start(Database.BatchableContext batchableContext) {
-        // Each batchable method runs in a separate transaction,
-        // so store the first transaction ID to later relate the other transactions
-        this.originalTransactionId = Logger.getTransactionId();
+  public Database.QueryLocator start(Database.BatchableContext batchableContext) {
+    // Each batchable method runs in a separate transaction,
+    // so store the first transaction ID to later relate the other transactions
+    this.originalTransactionId = Logger.getTransactionId();
 
-        Logger.info('Starting BatchableLoggerExample');
-        Logger.saveLog();
+    Logger.info('Starting BatchableLoggerExample');
+    Logger.saveLog();
 
-        // Just as an example, query all accounts
-        return Database.getQueryLocator([SELECT Id, Name, RecordTypeId FROM Account]);
+    // Just as an example, query all accounts
+    return Database.getQueryLocator([SELECT Id, Name, RecordTypeId FROM Account]);
+  }
+
+  public void execute(Database.BatchableContext batchableContext, List<Account> scope) {
+    // One-time call (per transaction) to set the parent log
+    Logger.setParentLogTransactionId(this.originalTransactionId);
+
+    for (Account account : scope) {
+      // Add your batch job's logic here
+
+      // Then log the result
+      Logger.info('Processed an account record', account);
     }
 
-    public void execute(Database.BatchableContext batchableContext, List<Account> scope) {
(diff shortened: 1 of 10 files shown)


jongpie:
@jamessimone this is great! I've got some thoughts below on some of the specifics, but I think we should definitely get this added in.

> potentially also creating a log entry for async context when it's first set (possibly always logging? or also reporting on when it's been called with the current context already set?)

I've struggled a bit with deciding in `Logger.cls` when to use `System.debug()` vs on of the logging methods - with the logging methods, I worry about Nebula Logger auto-generating too many extra `LogEntry__c` records that eat up space & people might not want them... But in this case, I do think it's helpful for an entry to be auto-created when the async context is set, so using the `info()` logging method seems better than the `System.debug()` statement I had before. For most situations, the async contexts will only be set once per transaction (so "only" 1 more `LogEntry__c` per async transaction). And being able to see explicitly when the context is set I think could be a useful piece of info to have to fully understand what's going on in an async transaction.
  - I feel the same way about the scenario methods in `Logger.cls`, like `Logger.setScenario()`. Right now, I use some `System.debug()` methods because I wanted to avoid generating more `LogEntry__c` records - but now I'm thinking auto-generating entries (using `info()` instead of `System.debug()`) would be really helpful for troubleshooting/understanding when a scenario changes in a transaction.

> appending some additional properties to the Logger.AsyncContext inner class to capture more data about finalizers. This leads to an INFO message like such:

I love the new properties & the resulting JSON, that seems really nice to have the finalizer details added. A few specific thoughts I had:
- I completely agree about the 'External entry point' stack trace, it's not great, but I like the idea of leaving it so people can at least still see what stack trace the platform provides
- For the `finalizerException` property, you've currently have it as `public final String finalizerException;` and set with the message + stack trace string (`this.finalizerException = finalizerContext.getException()?.getMessage() + '\n' + finalizerContext.getException()?.getStackTraceString() ?? null;`). If we made it an instance of `Exception` instead, we could also call `setExceptionDetails()` on the `info()` builder you added, which would then set all of the `LogEntry__c.Exception*` fields. Since the stack trace string is just 'External entry point', we may not get a lot of good info out of it, but it still seems like a good re-use of `setExceptionDetails()` and the existing `Exception` fields. Something like
    ```apex
    info('Nebula Logger - Async Context: ' + System.JSON.serializePretty(asyncContext, true)).setExceptionDetails(asyncContext.finalizerException);
    ```

jongpie:
@jamessimone I pushed [a commit](https://github.com/jongpie/NebulaLogger/pull/787/commits/3366d40d5b3d89d043303640153e81ac8719a73f) to add tests for the new `INFO` entries that are now auto-generated. Adding those tests also brought up the issue that `Exception` isn't serializable (something I completely forgot about), so I made a small change in `Logger` (using a `Map)` to work around that when the `AsyncContext` object is serialized & written to the `Message__c` field. Whenever you have a chance to take a look, let me know what you think.

jamessimone:
Looks good to me!

jongpie (approved):
Thanks @jamessimone! If you'd like to handle merging this & publishing release notes, feel free to do so - or I'm happy to handle it tomorrow.

jamessimone:
I'll take care of it

PR #789, opened 2024-10-27
Added more details in logger LWC's console statements

jongpie:
When using the `logger` LWC and JavaScript console logging is enabled (via `LoggerSettings__c.IsJavaScriptConsoleLoggingEnabled__c`), Nebula Logger automatically calls functions in the browser's `console` so devs can easily see the generated log entry directly in their browser (instead of having to find it in `LogEntry__c`). This PR improves the component log entry JSON that's printed using `console` statements - the stringified object now includes more details, such as details about the logged error/exception (when an error is logged), any tags added to the entry, and the user's localized version of the entry's `timestamp`.

- This includes a change to how Nebula Logger internally calls the browser's `console` functions: now, `console` functions are called with a 1 second delay (using `setTimeout()`) so that additional details can be added to the log entry (using the builder methods) before the log entry is stringified & printed out. This may not be a perfect solution, but seems to work in most cases of using the JS builder methods.

To show the difference in the changes, this JavaScript was used to generate some example log entries:
 ```javascript
this.logger.setScenario('Some demo scenario')
const someError = new TypeError('oops');
this.logger.error('Hello, world!')
    .setExceptionDetails(someError)
    .setField({ SomeLogEntryField__c: 'some text from loggerLWCGetLoggerImportDemo' })
    .addTags(['Tag-one', 'Another tag here']);
```

In the screenshot below, 2 versions of the output are shown:
1. **Before**:
    - No details are included in the output about the error, the tags, or the custom field mappings.
    - The timestamp is only shown in UTC.
2. **After**:
    - The output now includes details about any errors/exceptions logged using the builder function `setExceptionDetails()`.
        - Previously the output only included the logged message - _no_ details were included about the logged exception/error, which made it _much_ harder to troubleshoot whatever the relevant error was 🙃 
        - The error output includes details about the metadata source, making it easy to tell if the error was caused by another LWC, an Apex controller method, or an Apex trigger.
    - When devs use any of Nebula Logger's optional JS features (scenarios, tags, and custom field mappings), the details are automatically included in the `console` statements' output
     - For devs that do not leverage these optional features, the relevant properties are automatically excluded from the `console` output. This keeps the output slimmer & easier to read.
     - Each entry's timestamp is now shown both in UTC time (using `new Date().toISOString()`) and in the user's local format (using `new Date().toLocaleString()`). This makes it a little easier for devs to troubleshoot exactly when an entry was generated in their browser.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,13 @@ The most robust observability solution for Salesforce experts. Built 100% native
 
 ## Unlocked Package - v4.14.17
 
-[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015ocHQAQ)
-[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015ocHQAQ)
+[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015ocRQAQ)
+[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015ocRQAQ)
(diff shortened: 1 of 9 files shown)


jongpie:
@jamessimone if you have some time this week, I'd love to get your thoughts on these changes. I sent you an early preview of some of it a few days ago, but I've made several more tweaks to the output since then.

PR #840, opened 2025-02-17
Fix Stacktrace Parsing & Reduce logger.test.js

### LINKED ISSUE #780: logger.test.js is a bit of a mess
> One possibility that might help to clean things up in that test class such that you don't need to repeat yourself so much would be using parameterized tests:
> 
> ```js
> const getMarkupLogger = async () => {
>   const logger = createElement('c-logger', { is: Logger });
>   document.body.appendChild(logger);
>   await flushPromises();
>   return logger;
> };
> 
> // and then in a describe block
>   it.each([[createLogger], [getLogger], [getMarkupLogger]])('returns user settings', async loggingFunction => {
>     getSettings.mockResolvedValue({ ...MOCK_GET_SETTINGS });
>     const logger = await loggingFunction();
> 
>     const userSettings = logger.getUserSettings();
> 
>     expect(userSettings.defaultSaveMethod).toEqual('EVENT_BUS');
>     expect(userSettings.isEnabled).toEqual(true);
>     expect(userSettings.isConsoleLoggingEnabled).toEqual(false);
>     expect(userSettings.isLightningLoggerEnabled).toEqual(false);
>   });
> ``` 
> 
> In this example, each inner list passed to `.each` can have any number of arguments that are then passed to the second function, a la:
> 
> ```js
>  it.each([
>     [createLogger, 'createLogger'],
>     [getLogger, 'getLogger deprecated'],
>     [getMarkupLogger, 'markup logger']
>   ])('returns user settings', async (loggingFunction, text) => {
>  // ...
> })
> ```
> 
> What do you think? I will continue reviewing, but that might really help to cut down on the duplication in that file

_Originally posted by @jamessimone in https://github.com/jongpie/NebulaLogger/issues/775#issuecomment-2385646460_

_Follow-up comment posted by @jongpie in https://github.com/jongpie/NebulaLogger/pull/775#issuecomment-2386106060_

jamessimone:
* Fixes #780 by cleaning up `logger.test.js`
* Fixes an issue in `LoggerStackTrace.js` that would cause an unintended exception when there isn't a valid JS stacktrace

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,11 @@ The most robust observability solution for Salesforce experts. Built 100% native
 
 ## Unlocked Package - v4.15.7
 
-[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pEdQAI)
-[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pEdQAI)
+[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pEnQAI)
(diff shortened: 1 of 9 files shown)


jongpie (approved):
Soooooo many deleted lines - it's beautiful!

PR #849, opened 2025-03-01
Custom Index for Log__c.ParentLogTransactionId__c

### LINKED ISSUE #847: Add Index to Log__c.ParentLogTransactionId__c
### New Feature Summary

We are using Nebula to log API requests coming into a custom REST endpoint.  We are seeing some performance issues in a query (LogManagementDataSelector.getLogsWithoutParentLogByParentTransactionId: line 273 that is using a lot of resources.

We were seeing performance degradation and throttling enforcement (by Salesforce) when we hit a spike in the endpoint usage.  One of the less performant queries that was contributing to the issue.

Query: SELECT Id, ParentLogTransactionId__c, ParentLog__c, ParentLog__r.TransactionId__c, TransactionId__c FROM Log__c WHERE (ParentLogTransactionId__c IN :tempVar1 AND ParentLog__c = NULL)

Salesforce suggested adding an Index to the ParentLogTransactionId__c field.  I wasn't sure if this was possible to do from the package side, but it appears if you add it to a Dev Org (with Salesforce's help), you can include it in the package.  Otherwise customers will need to add these (and any other indexes) on their own.

Ref: Salesforce Case#469926652

Usage:

public Object doPost() {
  RestRequest request = RestContext.request;
  Logger.info(request.requestURI).setRestRequestDetails(request);
  //.. do stuff
  Logger.saveLog();
  return response;
}

jamessimone:
Fixes #847 by adding Log__c.ParentLogTransactionId__c custom index

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 The most robust observability solution for Salesforce experts. Built 100% natively on the platform, and designed to work seamlessly with Apex, Lightning Components, Flow, OmniStudio, and integrations.
 
-## Unlocked Package - v4.15.7
+## Unlocked Package - v4.15.8
 
 [![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pEnQAI)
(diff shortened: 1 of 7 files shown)


jongpie (approved):
@jamessimone this looks great to me, thanks so much for taking care of this! Feel free to merge & release this if you'd like, or I can do it this week.

PR #857, opened 2025-03-26
Bugfix: entry save override is not always used

### LINKED ISSUE #856: LogEntryEventHandler.filterLogEntryEventsToSave filters out entries that were forcefully saved with Logger.newEntry(String, Boolean)
@jongpie  – I believed I had tested this, but the developers noticed that although newEntry's shouldSave method allows inserting into the log's buffer by bypassing the log level, the platform event subscriber appears to run another check in filterLogEntryEventsToSave, ensuring the logging level matches. As a result, the log entry isn’t actually saved. Is this expected? If so, are there any other options to persist the entry regardless of the log level setting?

_Originally posted by @jvega-intakedesk in https://github.com/jongpie/NebulaLogger/discussions/855#discussioncomment-12603857_

jongpie:
Fixed #856 by adding a few code & config changes to ensure that entries are correctly saved when developers have overridden the `shouldSave` flag on entries when calling one of the `Logger.newEntry()` method overloads.

- Added new fields `EntrySaveReason__c` on `LogEntryEvent__e` & `LogEntry__c` to explicitly track why an entry is saved, with 2 possible values:
    1. `Logging Level Met`: indicates that the entry was saved because the entry's specified logging level meets the logging user's configured logging level settings, stored in `LoggerSettings__c.LoggingLevel__c`.
    2. `Override`: indicates that the entry was saved because a developer has overridden in code the `shouldSave` flag for the entry. This is done by developers to ensure certain entries are always logged, regardless of the user's configured logging level.
- Updated `LogEntryEventBuilder` and `LogEntryEventHandler` to set the new fields on `LogEntryEvent__e` and `LogEntry__c`.
    - Going forward, all new `LogEntryEvent__e` and `LogEntry__c` records will now have a value populated in `EntrySaveReason__c`.
    - Existing `LogEntry__c` records will still have `null` - no changes are being introduced to backfill the data.
- Updated the logic in `LogEntryEventHandler.filterLogEntryEventsToSave()` to check the value of the new field `LogEntryEvent__e.EntrySaveReason__c`. Any entries with a save reason of `Override` are always saved in `LogEntry__c` (which is the piece that directly fixes #856).
- The flexipage `LogEntryRecordPage` has also been updated to display the new field `LogEntry__c.EntrySaveReason__c` when it's populated. For any existing `LogEntry__c` records with a `null` value, the flexipage will hide the field.

    image

    image

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 The most robust observability solution for Salesforce experts. Built 100% natively on the platform, and designed to work seamlessly with Apex, Lightning Components, Flow, OmniStudio, and integrations.
 
-## Unlocked Package - v4.15.8
+## Unlocked Package - v4.15.9
 
 [![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pFCQAY)
 [![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pFCQAY)
--- a/nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls
+++ b/nebula-logger/core/main/log-management/classes/LogEntryEventHandler.cls
@@ -85,8 +85,11 @@ public without sharing class LogEntryEventHandler extends LoggerSObjectHandler {
         continue;
       }
 
-      if (String.isBlank(loggingUserSettings.DefaultPlatformEventStorageLoggingLevel__c)) {
-        // DefaultPlatformEventStorageLoggingLevel__c is optional - if it's null, then always save the event
+      // DefaultPlatformEventStorageLoggingLevel__c is optional - if it's null, then always save the event
+      if (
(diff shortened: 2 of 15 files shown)


jvega-intakedesk (approved):
It’s very similar to what we did as a temporary workaround. We avoided changing the core logic by cloning the LogEntryEventHandler and updated the handler to our clone class. Thanks for the quick turnaround.

jongpie:
@jvega-intakedesk that's great to hear that you took a similar approach, and your workaround of cloning `LogEntryEventHandler` sounds like a great way to handle this in the meantime. Would it be possible for you & your team to install a beta version of this PR in a sandbox (or scratch org) to help test it out? If not, no worries, but it'd be helpful if you're able to try this out before I merge/release this.

If you're able to test this, here are the (beta) package installation details. I would only try this in a **non-critical** sandbox (or a scratch org, if you use scratch orgs).

- **CLI install option:** `sf package install --wait 20 --security-type AdminsOnly --package 04t5Y0000015pG5QA`
- **UI install option:** https://test.salesforce.com/packaging/installPackage.apexp?p0=04t5Y0000015pG5QAI

jvega-intakedesk:
@jongpie - The developer team confirmed it's now working as expected.

jongpie:
@jvega-intakedesk that's great to hear, thanks for helping to test it! I'll try to merge this ASAP.

PR #884, opened 2025-07-22
Bugfix: logger LWC Initialization Message

### LINKED ISSUE #870: Disabling system messages in LWC component
Hello, I want to use NebulaLogger for errors in public site, but I don't want initialization and error messages in browser's console.

On c/logger/loggerService.js file I can see functions to disable messages:
```js
let areSystemMessagesEnabled = true;

export function enableSystemMessages() {
  areSystemMessagesEnabled = true;
}

export function disableSystemMessages() {
  areSystemMessagesEnabled = false;
}
```
But i'm not sure how to use these as I can only import c/logger and not c/logger/loggerService.js. Is it possible to provide it as a setting when initializing logger in LWC?

jongpie:
Fixed #870 by updating the `logger` LWC initialization message to only print when JavaScript console logging is enabled via `LoggerSettings__c`. Also resolved some test isolation issues in `logger.test.js` + test improvements + code cleanup

- Updated `loggerService.js` to only call `console.info()` with an init message ('logger component initialized') when console logging is enabled via `LoggerSettings__c.IsJavaScriptConsoleLoggingEnabled__c`
- Replaced `enableSystemMessages()` and `disableSystemMessages()` in `loggerService.js` with a new static variable `LoggerService.areSystemMessagesEnabled` (intended for internal-use only)
- Updated `logger.test.js` to reset `LoggerService.hasInitialized` and `LoggerService.areSystemMessagesEnabled` static variables in `beforeEach()` (instead of `afterAll()`)
- Switched to using `jest.resetAllMocks()` instead of `jest.clearAllMocks()` in `logger.test.js` for better test isolation
- Updated all test functions to consistently use all 3 logger creation methods (`createLogger`, `getLogger`, and `getMarkupLogger`)

Also include a 🤏 liiiiiiitle bit of scope creep:
- Updated all Jest tests to import createElement from '@lwc/engine-dom' (instead of 'lwc')
- Updated the project's threshold for code coverage in `codecov.yml` to compensate for the lower code coverage in LWC tests for now (currently around ~86%). Eventually, tests & code coverage for LWC will be improved - but for now, the constant pipeline alerts from codecov.io aren't particularly helpful 😅 
- Cleaned up a small code block in `LogBatchPurger` using the null coalescing operator (??) to reduce the lines of code needed
- Corrected the label on field `LoggerScenarioRule__mdt.IsLogAssignmentEnabled__c` to use the term `Logger Scenario`, instead of the deprecated term `Log Scenario`

### DIFF
--- a/nebula-logger/core/main/logger-engine/lwc/logger/__tests__/logger.test.js
+++ b/nebula-logger/core/main/logger-engine/lwc/logger/__tests__/logger.test.js
@@ -1,7 +1,6 @@
-import { createElement } from 'lwc';
+import { createElement } from '@lwc/engine-dom';
 import FORM_FACTOR from '@salesforce/client/formFactor';
-import { BrowserContext } from '../loggerService';
-import LoggerService from '../loggerService';
+import LoggerService, { BrowserContext } from '../loggerService';
 // Recommended import getLogger & deprecated import createLogger
 import { getLogger, createLogger } from 'c/logger';
 // Legacy markup-based approach
--- a/nebula-logger/core/main/logger-engine/lwc/logger/__tests__/logger.test.js
+++ b/nebula-logger/core/main/logger-engine/lwc/logger/__tests__/logger.test.js
@@ -1,6 +1,7 @@
 import { createElement } from 'lwc';
 import FORM_FACTOR from '@salesforce/client/formFactor';
-import { BrowserContext, enableSystemMessages, disableSystemMessages } from '../loggerService';
+import { BrowserContext } from '../loggerService';
+import LoggerService from '../loggerService';
 // Recommended import getLogger & deprecated import createLogger
 import { getLogger, createLogger } from 'c/logger';
 // Legacy markup-based approach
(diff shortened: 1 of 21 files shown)


nebula-logger/core/main/logger-engine/lwc/logger/__tests__/logger.test.js
```diff
@@ -1,6 +1,7 @@
 import { createElement } from 'lwc';
 import FORM_FACTOR from '@salesforce/client/formFactor';
-import { BrowserContext, enableSystemMessages, disableSystemMessages } from '../loggerService';
+import { BrowserContext } from '../loggerService';
+import LoggerService from '../loggerService';
```
jongpie, line 4: @jamessimone random readability/style-preference question for you - for imports like this (with a default + named exports), do you go for a one-liner? Or do you like to keep them separated?

```suggestion
import LoggerService, { BrowserContext } from '../loggerService';
```

Issue #903, opened 2025-09-05
Add support for Database Result parameters in CallableLogger


wherestheguac:
### New Feature Summary

Hi, I've been using the CallableLogger interface to work with NebulaLogger, but noticed that it isn't possible to pass any Database Save/Delete/Etc Result instances through in the argument map. Passing them through as a string after serializing them is possible but it would be great to have this be available by default in the Callable interface if possible.

Thanks for making this package available!

PR #905, opened 2025-09-10
Add Database Result Support to CallableLogger class

wherestheguac:
I took a stab at adding support for Database Result classes in `CallableLogger`. I added parameters to take in single result objects as well as a list, and added a parameter to pass through the type of result. I didn't bother adding support to pass through a recordId since that can already be accomplished through the 'recordId' parameter.

I ended up adding two generic implementations of `LogEntryEventBuilder.setDatabaseResult` to allow the log builder to take in a generic list of type `Object`, or a single `Object` instance. I went with this vs trying to route the result(s) through the overloaded methods because there isn't really a way to cast in Apex using the `System.Type` that I'm aware of, and `LogEntryEventBuilder.setDatabaseDetails` is already abstracted to take in a `List<Object>` parameter anyway. It's not exposed through the `Logger` class so I figured it would be fine.

Currently the `'resultType'` parameter would not be required, since there is a call: `resultType.getName();` in `LogEntryEventBuilder.setDatabaseDetails` that I switched to `resultType?.getName();` to handle the case where the `'resultType'` parameter is not passed to CallableLogger.

I also added methods to `LoggerMockDataCreator` to create lists of mocked Database Result objects and updated `Logger_Tests` to use those methods as well.

I didn't update the `Nebula Logger Demo` Omniscript since the other record parameters weren't in there either.

Tests were passing when I ran through the github build actions locally but I'll update the branch if anything fails.

Let me know if there is anything else that needs to be taken into consideration here.

### DIFF
--- a/nebula-logger/core/tests/logger-engine/classes/CallableLogger_Tests.cls
+++ b/nebula-logger/core/tests/logger-engine/classes/CallableLogger_Tests.cls
@@ -592,6 +592,35 @@ private class CallableLogger_Tests {
   }
 
   @IsTest
+  static void it_adds_new_entry_with_delete_result_and_null_type_when_using_standard_approach() {
+    System.LoggingLevel loggingLevel = System.LoggingLevel.ERROR;
+    String message = 'some log entry message';
+    Database.DeleteResult result = LoggerMockDataCreator.createDatabaseDeleteResult(true);
+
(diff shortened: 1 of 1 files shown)


wherestheguac:
Related issue: https://github.com/jongpie/NebulaLogger/issues/903

Issue #922, opened 2025-11-13
Logger test hardcodes value for custom field and tests are failing


mbrezina:
### Package Edition of Nebula Logger

Unlocked Package

### Package Version of Nebula Logger

v4.16.5

### New Bug Summary

We are using Nebula for logging of errors in a Salesforce org. We added a field to Log_Entry__c object and updated event and the mapping in custom metadata. However we use a Nebula package in scratch orgs together with other custom packages and we run all tests. I reproduced the same test failure in my scratch org where I have Nebula installed, I ran all tests in the org. We were getting this error: 

Class.LogEntryEventHandler_Tests.it_should_not_create_log_entry_data_when_platform_event_storage_logging_level_is_not_met: line 88, column 1 
LogEntryEventHandler_Tests.it_should_not_normalize_scenario_data_when_data_normalization_is_disabled Fail System.DmlException: Upsert failed. First exception on row 0; first error: INVALID_OR_NULL_FOR_RESTRICTED_PICKLIST, Process Name: bad value for restricted picklist field: ProcessName__c Test STRING: [ProcessName__c] 
Class.LoggerDataStore.Database.upsertRecords: line 276, column 1 
Class.LogEntryEventHandler.upsertLogEntries: line 410, column 1 
Class.LogEntryEventHandler.executeAfterInsert: line 60, column 1 
Class.LoggerSObjectHandler.execute: line 125, column 1 
Class.LoggerMockDataStore.MockEventBus.deliver: line 157, column 1

After troubleshooting and searching in Salesforce and in Nebula repository I discovered it is in this class - see the line 896, it hardcodes the string field value to be [Name of the custom field ] Test STRING. In our case we use a global value set which is a restricted picklist by default so the test is failing. Our field is a picklist value, not string but still it works as String. I haven´t found this value hardcoded anywhere else in the code in the repository:  

[NebulaLogger/nebula-logger/core/tests/configuration/utilities/LoggerMockDataCreator.cls at v4.16.5 · jongpie/NebulaLogger · GitHub](https://url.uk.m.mimecastprotect.com/s/zVOWCRoy1CmMqYGys9fLC1dXT0?domain=github.com)  

The custom field is ProcessName__c/Process Name. We were guessing and are added various values to the global set, finally this was the correct combination, then the tests passed:

```
<customValue>
    <fullName>ProcessName__c Test STRING</fullName>
    <default>false</default>
    <label>ProcessName__c Test STRING</label>
</customValue>
[...]

PR #925, opened 2025-11-20
v4.17.0 - Summer '25 Release (API v64.0)

### LINKED ISSUE #777: Code line highlighting in Prism JS doesn't always appear
For some reason, the line highlighting used for Apex code snippets in `c/loggerCodeViewer` (powered by Prism JS) does not always correctly show up. 

- For `LogEntry__c` records without an exception, only 1 instance of the component is shown to display origin details. In this case, the line highlighting seems to never really work
  
  image

- For `LogEntry__c` records with an exception, 2 instances of the component are - one for the origin details, and one for the exception details. Typically, these records show the highlighting on both components, but even this is sometimes flaky
  
  image

My theory is that this is happening because of how/when the PrismJS static resource is loaded, but so far, I haven't been able to fix it.

### LINKED ISSUE #881: LWC embedded in record page not logging component information or exception details correctly

Unlocked Package

4.16.2

### New Bug Summary

I'm using Nebula Logger in a LWC embedded in a Lightning Record Page. Logs do generate, but they don't seem to be "correct" in the sense that I don't seem to be seeing any information logged into the "Lightning Component Information" or "Exception Information" sections like the ReadMe shows I should be.

How I'm using it:

`import { getLogger } from "c/logger";
logger = getLogger();
this.logger.error("Error in Component Name: " + JSON.stringify(error)).setExceptionDetails(error);
this.logger.saveLog();`

An example of the exception object being passed in `setExceptionDetails` is:

`
{
    "status": 400,
    "body": {
        "message": "An error occurred while trying to update the record. Please try again.",
        "statusCode": 400,
        "enhancedErrorType": "RecordError",
        "output": {
            "errors": [
                {
                    "constituentField": null,
                    "duplicateRecordError": null,
                    "errorCode": "FIELD_CUSTOM_VALIDATION_EXCEPTION",
                    "field": null,
                    "fieldLabel": null,
                    "message": "If you're moving the Lead to an Unqualified Status you must select an Unqualified Reason!"
                }
            ],
            "fieldErrors": {}
        },
        "id": "-163148827"
    },
    "headers": {},
    "ok": false,
    "statusText": "Bad Request",
    "errorType": "fetchResponse"
}
`

This is all I end up getting log wise currently:

Image

Logs that do come through have Origin Source API Name comes through as:

`https://redacted.sandbox.lightning.force.com/lightning/r/Lead/LeadId/modules/c/enhancedPath`

and Stack Trace as:

`at EnhancedPath._logErrorAndToast (https://redacted.sandbox.lightning.force.com/lightning/r/Lead/LeadId/modules/c/enhancedPath.js:671:21)
at eval (https://redacted.sandbox.lightning.force.com/lightning/r/Lead/LeadId/modules/c/enhancedPath.js:646:18)`

### LINKED ISSUE #910: CallableLogger_Tests: System.UnexpectedException

Unlocked Package

v4.16.5

### New Bug Summary

The above test class seems to be failing intermittently. There doesnt seem to be any good rational to determine why we see these issues.  

**it_adds_new_entry_with_tags_when_using_standard_approach**

```
System.UnexpectedException: class java.lang.String cannot be cast to class java.lang.Integer (java.lang.String and java.lang.Integer are in module java.base of loader 'bootstrap')
Stack Trace: External entry point
```

any help understanding this would be greatly appreciated

jongpie:

## Summer '25 Release Upgrade

- Bumped all metadata to API `v64.0` (Spring '26 release)
  - Also updated the list of picklist values in several 'API version' picklist fields on `Log__c` and `LogEntry__c`

## Apex Logging Bugfixes

- Updated `Logger` to more accurately reflect the current quiddity value when logging the system message `Logger - Saving X log entries via <quiddity>`
- Fixed #910 - Updated `CallableLogger` to use `List<String>` instead of `List<Object>` for tags passed in via OmniStudio
  - Previously using `List<Object>` caused some exceptions, and the values should always be strings anyway

## JavaScript Logging Bugfixes

- Updated `logger` LWC to reflect the latest changes in Salesforce's JavaScript stack traces
  - These changes broke the previous stack trace parsing logic - hopefully the modified logic is more resilient to future changes
  - The Jest tests in `logger.test.js` have also been expanded to include parsing examples of stack traces from Chrome, Firefox, Edge - with & without debug mode (which can make a difference in the stack traces & parsing logic)
- Updated `logger` LWC to immediately instantiate an internal instance of `LoggerService`
  - This was previously done in `connectedCallback`, which added an extra unnecessary event cycle that could cause some occasional goofiness

## PrismJS Enhancements & Bugfixes

- Upgraded the version of PrismJS in `LoggerResources` static resource from `1.29.0` to `1.30.0`
- Updated some CSS in `loggerCodeViewer.css` to correct some rendering issues due to changes in SLDS + SLDS v2
- Fixed #777 - Corrected an issue in `loggerCodeViewer` that caused PrismJS's line highlighting to not work correctly when only 1 instance of `loggerCodeViewer` (used by other LWCs like `logEntryMetadataViewer`) was on a page

## LWC Testing Improvements

- Fixed #881 - Added more Jest tests in to the LWCs tests below to validate existing functionality that wasn't previously covered by tests:
  - `logEntryEventStream.test.js`
  - `logEntryMetadataViewer.test.js` 
  - `loggerHomeHeader.test.js`
  - `logger.test.js`
  - `loggerPageSection.test.js`
  - `loggerSettings.test.js`
  - `logViewer.test.js`
- Added a `recipes` LWC `loggerWrapper` to mimic orgs that use Nebula Logger's `logger` LWC through their own logging LWC

## Apex Testing Improvements

- Updated `LogEntryEventHandler_Tests` to only use mock CMDT records (instead of relying on the metadata records deployed to the org)
  - All other test classes were previously updated to only use mock records, but this particular test class was overlooked
- Partially addressed #922 - Updated pipeline-only tests in `LogEntryEventHandler_Tests_FieldMappings.cls` to validate that custom picklist fields with restricted values are handled properly

# Pipeline Changes

- Updated scratch org definition files to tweak how OmniStudio is enabled in scratch orgs
- Updated pipeline & local dev scripts to use OmniStudio managed package v258.6 (Winter '26 release)
- Upgraded some dev dependencies in `package.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,11 @@ The most robust observability solution for Salesforce experts. Built 100% native
 
 ## Managed Package - v4.17.0
 
-[![Install Managed Package in a Sandbox](./images/btn-install-managed-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?mgd=true&p0=TODO_MGD_PKG)
-[![Install Managed Package in Production](./images/btn-install-managed-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?mgd=true&p0=TODO_MGD_PKG)
+[![Install Managed Package in a Sandbox](./images/btn-install-managed-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?mgd=true&p0=04tg70000000r5xAAA)
+[![Install Managed Package in Production](./images/btn-install-managed-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?mgd=true&p0=04tg70000000r5xAAA)
 [![View Milestone](./images/btn-view-managed-package-milestone.png)](https://github.com/jongpie/NebulaLogger/milestone/16?closed=1)
 
-`sf package install --wait 30 --security-type AdminsOnly --package TODO_MGD_PKG`
+`sf package install --wait 30 --security-type AdminsOnly --package 04tg70000000r5xAAA`
 
 ---
 
--- a/nebula-logger/managed-package/sfdx-project.json
+++ b/nebula-logger/managed-package/sfdx-project.json
@@ -35,6 +35,6 @@
     "Nebula Logger - Managed Package@4.14.0-summer-24-release": "04t5Y0000015oPqQAI",
     "Nebula Logger - Managed Package@4.15.0-winter-25-release": "04t5Y0000015odFQAQ",
     "Nebula Logger - Managed Package@4.16.0-spring-25-release": "04t5Y0000015pGtQAI",
-    "Nebula Logger - Managed Package@4.17.0-summer-25-release": "04tKe000000Pbr7IAC"
+    "Nebula Logger - Managed Package@4.17.0-summer-25-release": "04tg70000000r5xAAA"
   }
 }


jamessimone:
I love it when a plan comes together!

PR #959, opened 2026-03-03
Fix: Chunked regex with overlapping Matcher for boundary-safe data masking

### LINKED ISSUE #639: System.LimitException: Regex too complicated from LogEntryEventBuilder.applyDataMaskRules:

Unlocked Package

v4.12

### New Bug Summary

I get an expected error when running a batch + queuable job due to locking error from a Query. So I have Nebula Logger that I want to log this error, but when it's trying to log my original error, I get an exception "System.LimitException: Regex too complicated". 

Stack Trace from Nebula Logger (error is LogEntryEventBuilder.applyDataMaskRules): 
Class.LogEntryEventBuilder.applyDataMaskRules: line 928, column 1
Class.LogEntryEventBuilder.setMessage: line 151, column 1
Class.Logger.error: line 838, column 1

anthonygiuliano:
## Summary

Fixes #639 — `System.LimitException: Regex too complicated` when `applyDataMaskRules` processes long strings (~35K+ chars).

The existing chunking logic in `applyDataMaskRuleToLongLine` had three correctness issues that this PR addresses:

- **Longest-match deduplication**: When two overlapping chunks find a match at the same start position, the old code kept whichever was found first (shorter context). Now keeps the longest match, since the chunk with more trailing context produces the most accurate result.
- **Sorted left-to-right processing**: The old code processed matches in discovery order across chunks, which isn't guaranteed to be left-to-right. Now explicitly sorts start positions and skips matches consumed by previous replacements.
- **Single-pass `expandReplacement`**: The old iterative `String.replace('$' + i, groups[i])` reinterpreted `$N` patterns inside captured group values (e.g., if group 1 captured `PRICE:$3`, the `$3` would be replaced with group 3's value). The new implementation scans the replacement template left-to-right, only interpreting `$N` in the template itself — matching Java's `Matcher.appendReplacement` behavior.

## Test plan

- [x] 13 new targeted tests covering:
  - SSN in the overlap zone (no double-masking)
  - Multiple SSNs near the same chunk boundary
  - SSN at start/end of long strings
  - SSN exactly at the chunk step position
  - Credit card straddling a chunk boundary
  - Exact chunk-size and chunk-size+1 boundary conditions
  - Multiline where one line is exactly chunk size
  - Longer match wins when two chunks find same start
  - Left-to-right ordering across chunks (Map.keySet() has no guaranteed order in Apex)
  - Overlapping match skip (`start < pos` branch)
  - `$N` tokens inside captured group values are preserved literally (not re-expanded as group references)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -20,6 +20,11 @@ global with sharing class LogEntryEventBuilder {
   private static final String HTTP_HEADER_FORMAT = '{0}: {1}';
   private static final String NEW_LINE_DELIMITER = '\n';
 
+  @TestVisible
+  private static final Integer DATA_MASK_REGEX_CHUNK_SIZE = 4000;
+  @TestVisible
+  private static final Integer DATA_MASK_REGEX_OVERLAP_SIZE = 20;
```
jongpie: @anthonygiuliano could you share some insights on why these 2 values (4000 & 20) were chosen for the 2 constants? I think it's worth also adding some comment(s) here to explain the intent of these 2 constants & why these particular values are being used.

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -1150,12 +1155,108 @@ global with sharing class LogEntryEventBuilder {
...
+        dataInput = applyDataMaskRuleToChunkedText(dataInput, dataMaskRule.SensitiveDataRegEx__c, dataMaskRule.ReplacementRegEx__c);
       }
     }
     return dataInput;
   }
 
+  private static String applyDataMaskRuleToChunkedText(String text, String sensitiveDataRegEx, String replacementRegEx) {
+    if (text == null || text.length() <= DATA_MASK_REGEX_CHUNK_SIZE) {
+      return text == null ? text : text.replaceAll(sensitiveDataRegEx, replacementRegEx);
+    }
```
jongpie: I think this would read a little better if there were 2 separate/distinct `if` blocks:

```suggestion
    if (text == null) {
      return text null;
    }
    
    if (text.length() <= DATA_MASK_REGEX_CHUNK_SIZE) {
      return text.replaceAll(sensitiveDataRegEx, replacementRegEx);
    }
```

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -1150,12 +1155,108 @@ global with sharing class LogEntryEventBuilder {
...
+            result += replacement.substring(i, j);
+          }
+          i = j - 1; // -1 because the for loop increments
+          continue;
+        }
+      }
+      result += replacement.substring(i, i + 1);
+    }
+    return result;
+  }
```
jongpie: This method is also difficult to read/understand - there are a lot of loops with index variables being manipulated, and it's not clear what the intent is for a lot of the logic. Can we include some explanatory comments, along with examples of what this method is trying to do?

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
```diff
@@ -1150,12 +1155,108 @@ global with sharing class LogEntryEventBuilder {
...
+        if (!endByStart.containsKey(absStart) || absEnd > endByStart.get(absStart)) {
+          endByStart.put(absStart, absEnd);
+          List<String> groups = new List<String>();
+          for (Integer g = 0; g <= m.groupCount(); g++) {
+            groups.add(m.group(g));
+          }
+          groupsByStart.put(absStart, groups);
+        }
+      }
+    }
```
jongpie: This code is a bit difficult to understand why it's doing certain things. Can we include some explanatory comments, along with examples of what this method is trying to do?

jongpie (commented):
@anthonygiuliano thanks for working on this! Conceptually, these changes seem like they'll help with the data masking issues, but some of the specific code changes are a little difficult to understand. I added a few questions to the PR - would you mind taking a look whenever you have some time?

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
anthonygiuliano: Good question — digging in, the honest answer is there's no universal correct value, so I've added detailed comments on both constants and made them overridable.

**Overlap (20):** must be ≥ the longest value any enabled rule can match (else a value straddling a chunk boundary is missed). 20 covers the built-in rules (SSN ~11, credit card ~19 w/ separators). It can't be derived from a regex (max match length isn't generally computable), so for orgs with custom rules matching longer values it's overridable via an optional `LoggerParameter__mdt.DataMaskRegexOverlapSize` record.

**Chunk size (4000):** I benchmarked this (reproduced on a scratch org + a sandbox). With all shipped rules applied single-pass, the uncatchable `Regex too complicated` cliff is ~110K chars (realistic/diluted log shape) to ~220K (dense). It's a regex-engine *step* budget, not a char count — and notably the single worst rule (Mastercard's alternation + `\3` backref) sets the cliff; the number of rules doesn't. Chunk size turned out to be a *safety* knob, not a perf lever: CPU was flat across 1K–64K. So 4000 is deliberately conservative (~27× under the worst-case cliff) at zero perf cost, and also overridable via `LoggerParameter__mdt.DataMaskRegexChunkSize` for orgs whose custom rules push the failure point down. Full rationale + measurements are in the constant comments and the CMDT `Description__c` fields.

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
anthonygiuliano: Done — split into two distinct `if` blocks as suggested (used `return text;` for the null case).

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
anthonygiuliano: Addressed. Added a method docblock with a worked example (SSN straddling a chunk boundary) and inline comments explaining each step. The biggest offender for "index variables being manipulated" was `expandReplacement` — I replaced its hand-rolled `i`/`j` character scanner with a precompiled `Matcher` on `\$(\d+)`, which is much clearer and preserves the key property that a `$N` sequence *inside a captured value* is not re-expanded. Logic-equivalent and covered by existing + new tests.

nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls
anthonygiuliano: Added explanatory comments covering why chunks overlap, why matches are keyed by absolute start position, the longest-match dedup, the explicit sort (Apex `Map.keySet()` has no guaranteed iteration order), and the `start < pos` skip — with the worked example in the method docblock.

anthonygiuliano:
Thanks for the review. Summary of changes pushed:

- **Constants:** documented in detail + made overridable via `LoggerParameter__mdt` records (`DataMaskRegexOverlapSize`, `DataMaskRegexChunkSize`). Rationale: the right values depend on an org's custom rule match-lengths and payload shapes, so hard-coding wasn't appropriate — conservative measured defaults plus an escape hatch. The "why 4000 / why 20" detail is in the L26 thread.
- **Readability:** docblocks + worked examples on `applyDataMaskRuleToLongLine`; `expandReplacement` rewritten from the hand-rolled index scanner to a `Matcher` (logic-equivalent, covered by tests).
- **Tests:** added an overlap-override test; renamed `it_should_apply_data_mask_rule_to_digit_dense_string_without_limit_exception` → `it_should_mask_serialized_record_json_without_limit_exception` (the input is serialized-record JSON, and benchmarking showed dense input is actually the *safe* case — diluted/realistic text trips the limit sooner).
- All `LogEntryEventBuilder_Tests` pass (138).

Happy to split any of this out into a separate PR if you'd prefer to keep this one narrowly scoped to the comment/readability changes.

jongpie (approved):
@anthonygiuliano thanks again for working on this! I finally had a chance to spend more time reviewing the changes - the data masking changes themselves look great, I just did some cleanup & shuffled a few things around. Most notably, I made an inner class `DataMasker` to try to keep the relevant code all together, and moved a few things to the `LoggerParameter` class for consistency.

I'm working on release notes now - I'll merge this later today!

jongpie:
This is now available in [release `v4.19.2`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.19.2)!

PR #977, opened 2026-04-20
Adds new flag to relatedLogEntries LWC component

jamessimone:
*allowing flexipages and on the component itself filtering of log entries down to only entries associated with the record in question

image


jongpie (approved):
LGTM!!!! 👏 🥇 🥳 🎈

Issue #761, opened 2024-09-04
System.UserInfo.getSessionId() throws System.UnexpectedException


krishnandusarkarprax:
### Package Edition of Nebula Logger

Unlocked Package

### Package Version of Nebula Logger

4.13.17

### New Bug Summary

**Package Edition of Nebula Logger**
Unlocked Package

**Package Version of Nebula Logger**
4.13.17

**New Bug Summary**
When a user logs in through a server-to-server connected app with the setting "Issue JSON Web Token (JWT)-based access tokens for named users" enabled tries to use a `Logger.xxx` method, a Salesforce System Error is thrown.

Debugging reveals that the issue occurs when `System.UserInfo.getSessionId()` is called. Even with a try/catch block around this call, the code doesn't reach the catch block; instead, the error is thrown directly.

The only workaround is to disable the "Issue JSON Web Token (JWT)-based access tokens for named users" setting.

Defect #598 was reported a few months ago, leading to the resolution of adding a try/catch block around this code. However, despite the latest Salesforce Release (Winter '25) and Nebula Logger (version 4.13.17), the issue persists.

<img width="1512" alt="Screenshot 2024-09-04 at 10 30 33 PM" src="https://github.com/user-attachments/assets/f4ae600c-0ac2-41c5-ad68-a91f5dfc681d">
<img width="1512" alt="Screenshot 2024-09-04 at 10 31 05 PM" src="https://github.com/user-attachments/assets/fa32c081-37c2-4735-8129-d738cd60f7d2">

jongpie:
@krishnandusarkarprax thanks for reporting this! Trying to get the user's session has been such a headache for years (issues #97 is another example where this came up ~3 years ago), and I haven't run into this particular issue myself before, so I appreciate the info & context. 

Is it feasible for you to add this code snippet somewhere in your org (either in the `Logger` class somewhere, or one of your own Apex classes)? It just needs to be added in some Apex method that would be called in your scenario when you're using the setting "Issue JSON Web Token (JWT)-based access tokens for named users"

```apex
Integer currentSessionCount = [
    SELECT COUNT() FROM AuthSession
    WHERE UsersId = :System.UserInfo.getUserId() AND IsCurrent = TRUE
];
System.debug('Count of current sessions for the current user: ' + currentSessionCount);
```

[...]

PR #978, opened 2026-04-22
Bugfix: do not load Session Id in a JWT session (#761, #964)

mmablom:
Following my comment in #761 I tried some options.

The naive approach of querying the AuthSession directly seems contradictory to the optimizations done with the `LoggerEngineDataSelector`.

Following that I added the filtering to the AuthSession query in `LoggerEngineDataCollector` and also return the field `IsAssociatedWithJwtAccessToken`. When loading the Session Id we can check if the AuthSession is for JWT based on the cached proxy. 

One drawback is that this check does not work if 'QueryAuthSessionData' is set to false. For now I've landed on only skipping the Session Id load if we're definitely in a JWT session. I've also expiremented with always querying the AuthSession, but that would run counter to the intention of the parameter.

Another drawback is that this will invalidate the 'QueryAuthSessionDataSynchronously' parameter, since this will always query the AuthSession through the chain LOGGING_CONTEXT > transactionQuiddity > SESSION_ID.

For now I'll keep this a draft PR with the solution I landed on before working on test. @jongpie what do you think of this approach and/or do you see any other options to handle this issue?

### DIFF
--- a/nebula-logger/core/main/logger-engine/classes/Logger.cls
+++ b/nebula-logger/core/main/logger-engine/classes/Logger.cls
@@ -86,13 +86,44 @@ global with sharing class Logger {
     set;
   }
 
+  private static Boolean IS_JWT_SESSION {
+    get {
+      if (IS_JWT_SESSION == null) {
+        Boolean hasCurrentSession = false;
+        Boolean hasJwtSession = false;
+        
+        for (AuthSession session : [
+          SELECT IsCurrent, IsAssociatedWithJwtAccessToken
+          FROM AuthSession
+          WHERE UsersId = :System.UserInfo.getUserId()
+        ]) {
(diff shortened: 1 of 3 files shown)


jongpie:
@mmablom thanks for working on this! I'm reviewing your changes & looking into a few other ideas to see if there are any other ways to handle some of these issues. There may be a way to avoid some of the issues (without querying, etc.) - I'm working on setting up some scratch orgs to let me test out a few different scenarios, I'll let you know what I find.

jongpie:
@mmablom I'm closing out this PR, and I'm planning to (hopefully) go with the changes in PR #995 instead - it removes most of the usage of `System.UserInfo.getSessionId()` altogether. See [my comment here](https://github.com/jongpie/NebulaLogger/issues/761#issuecomment-4633778996) for more details.

PR #985, opened 2026-05-09
Adding Overloads to Logger.logDatabaseErrors() methods with List<SObjects>

### LINKED ISSUE #942: Add overloads to Logger.logDatabaseErrors() to accept List<SObject>
Currently, there are several overloads for `Logger.logDatabaseErrors()` to log only error instances of any of the various `Database` result classes. But the result classes themselves don't have any info about the actual record, other than possibly the record ID (which will be blank on `insert`).

To help give more context to the `Database` results, there should be overrides for `Logger.logDatabaseErrors()` to pass in the list of `SObject` records

- Assumption: the list of results & the list of records will be the same size & order (e.g., `results[0]` will be for `records[0]`, etc.)
- For each error result, find the matching record with the same index --> add it to a separate (temp) list
  - This is very similar to what's already done for any error results - only results with errors are added to a separate list `resultsToLog`, which is internally passed to `setDatabaseResult()`.
- Pass the list of matching records to `setRecord()` (alongside the existing call to `setDatabaseResult()`

DivyanshuBist:
PR for #942 ,
added overloads for Logger.logDatabaseErrors with parameter List<Sobject>


jongpie:
Thanks for making this PR @DivyanshuBist! I am planning to review & merge this in the next few days.

jongpie (approved):
Hi @DivyanshuBist - thanks for working on this PR! I spent the last day or two reviewing & have pushed several changes to the PR, including:

- Adding tests
- Changing the behavior when the list of database results & the list of SObject recoreds don't match. Now it just simply ignores the list of SObject records when there's a mismatch.
- **Scope creep**: there was some existing behavior in `LogEntryEventBuilder` that I didn't like, so I've tweaked the behavior in this PR

I'm planning to merge & release this later today 🥳

PR #992, opened 2026-05-28
Feature/889 skip system debug in managed package

### LINKED ISSUE #889: The managed package should never attempt to call `System.debug()`
In [the current version](https://github.com/jongpie/NebulaLogger/blob/v4.16.4/nebula-logger/core/main/logger-engine/classes/LogEntryEventBuilder.cls#L943) of `LogEntryEventBuilder`, it only checks the field `LoggerSettings__c.IsApexSystemDebugLoggingEnabled__c` to determine if it should call `System.debug()`. But for the managed package (any managed package), calling `System.debug()` won't work (Salesforce hides the managed package's debug lines from the customer org). This results in a performance tax, with no benefit to the managed package.

The builder class's code needs to be updated to also detect if the code is running in a namespace/managed package - if so, don't call `System.debug()`.

JMercie:
Added guard clause to prevent System.debug() being invoked on LogEntryEventBuilder when running as a managed package, taking advantage of already existing prop `NAMESPACE_PREFIX`. 

image

This close #889 

---

**Disclosure**: This PR contains AI generated code.


JMercie:
I'm wondering if we have to add a way to like globably disabled the system.debug for other classes like `Logger.cls` @jongpie.

PR #994, opened 2026-06-02
New Logger Parameter record 'OrganizationAllowlist' to control logging

jongpie:
Added a new `LoggerParameter__mdt` record, `OrganizationAllowlist`, to provide an optional way to control via custom metadata (CMDT) if logging should be enabled in the current org. This is useful in situations where sandboxes are created from a prod org (which clones prod's CMDT records), and orgs don't want logging to be enabled by default in new sandboxes.

There are 3 ways this record can be configured:

1.  **Out of the box (when `OrganizationAllowlist` is blank/empty)**: the `OrganizationAllowlist` record has no impact on logging / logging is fully controlled by `LoggerSettings__c` records.
2. **When the current org's 15 or 18 character ID is present**: logging is enabled based on `LoggerSettings__c.IsEnabled__c`.
3. **When the current org's ID is _NOT_ present**: logging is disabled for the whole org, regardless of what's been configured in `LoggerSettings__c.IsEnabled__c`.

> [!IMPORTANT]  
> TODO: right now, this means that if the `OrganizationAllowlist` has org IDs populated & the CMDT records are deployed to a scratch org, then logging will be disabled in the scratch org... and since scratch orgs are temporary environments, it's not feasible to add their org IDs. There may need to be slightly different behavior for scratch orgs - but determining that the current environment is a scratch org requires using a SOQL query on `Organization` 🤔

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,11 @@ The most robust observability solution for Salesforce experts. Built 100% native
 
 ## Unlocked Package - v4.18.4
 
-[![Install Unlocked Package in a Sandbox](./images/btn-install-unlocked-package-sandbox.png)](https://test.salesforce.com/packaging/installPackage.apexp?p0=04tg70000009GaDAAU)
-[![Install Unlocked Package in Production](./images/btn-install-unlocked-package-production.png)](https://login.salesforce.com/packaging/installPackage.apexp?p0=04tg70000009GaDAAU)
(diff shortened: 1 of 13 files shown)


jongpie:
I'm closing this PR without merging it - it was going to be a temporary workaround until I had time to work on some larger enhancements. But the approach in this PR has its own issues, so I've instead pivoted back to working on the larger enhancements (which I hope to release in the coming months).

PR #1021, opened 2026-08-24
Bugfix: Fixed NullPointerException on Flow Logging Actions

### LINKED ISSUE #997: Default Flow InputParameter leads to System.NullPointerException

Managed Package

v4.18

### New Bug Summary

This bug seems to have been introduced by salesforce via their UI change in Flows, which has a impact on the nebula logger that i cant seem to find a good fix for.

When i first open a action element in a flow that contains the nebula logger, it looks like this:
_(but all inputParameters set to **true**)_
Image

Here i am able to set which inputParamtere the flow should respect, which is good. 

However, after 1-2 seconds of load, it jumps to this:
Image

This removes the option to ignore some inputParameters, which leads to System.NullPointerException error when the flow is running that hits the logger action. _(as long as the inputParameters is set to true, where a value is not provided)_

Note: this is tested on 4.16 aswell, as i think salesforce introduced this UI change at the same times as 4.16 or 4.17.

jongpie:
Fixed #997 NullPointerException that occurs when Flow passes a null value for the optional inputs `Save Log` and `Throw FlowException for Fault Error Message`, which happens now that Salesforce has removed the per-input include toggles from the Flow Builder action panel 😐 

`tl;dr - Salesforce broke stuff (╯‵□′)╯︵┻━┻`


jongpie:
This has been released in [release `v4.19.5`](https://github.com/jongpie/NebulaLogger/releases/tag/v4.19.5)

